PR #158, opened 2014-12-16
Feature/new mail api

kalikaneko:
DO NOT MERGE
unfinished work.
pullreq only to be able to receive feedback.


kalikaneko:
merged also the refactor of incoming service

kalikaneko:
this is yet unfinished work, but I think some review of this huge changeset can already be made.
I'd like to merge something onto feature/async-api as soon as possible, and roll a pre-alpha as soon as we complete a minimally working workflow (send, receive, append to mailboxes).

kalikaneko:
force-pushed one more time, rebased against develop. note that this needs to be merged into feature/async api branch

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+
+    mboxwrapper_klass = MailboxWrapper
+
+    def __init__(self):
+        SoledadIndexMixin.__init__(self)
+
+    mboxwrapper_klass = MailboxWrapper
+
+    def __init__(self):
+        SoledadIndexMixin.__init__(self)
```
meskio: Two `__init__` a two `mboxwrapper_klass` with the same params and the same implementation? Is there something I'm missing or this is a copy-paste failure?

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        Get a DeferredLock that is unique for this subclass name.
+        Used to lock the access to indexes in the `get_or_create` call
+        for a particular DocumentWrapper.
+        """
+        return cls._k_locks[cls.__name__]
+
+    def __init__(self, **kwargs):
+        doc_id = kwargs.pop('doc_id', None)
+        self._doc_id = doc_id
+        self._future_doc_id = kwargs.pop('future_doc_id', None)
```
meskio: If `doc_id` and `future_doc_id` are arguments that must exist will not be more clear if they be explicit in the `__inti__` params?

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+    @property
+    def doc_id(self):
+        return self._doc_id
+
+    @property
+    def future_doc_id(self):
+        return self._future_doc_id
+
+    def set_future_doc_id(self, doc_id):
+        self._future_doc_id = doc_id
```
meskio: Not important at all, but will it be better to use `@future_doc_id.setter` decorator?

https://stackoverflow.com/questions/6618002/python-property-versus-getters-and-setters

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        Create the documents for this wrapper.
+        Since this method will not check for duplication, the
+        responsibility of avoiding duplicates is left to the caller.
+
+        You might be interested in using `get_or_create` classmethod
+        instead (that's the preferred way of creating documents from
+        the wrapper object).
+
+        :return: a deferred that will fire when the underlying
+                 Soledad document has been created.
```
meskio: No documentation on the possible failures that the Deferred can fire. I'm not sure how much we should enforce this documentation, but might help to remember to add errbacks and don't get hard to debug problems.

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        def return_wrapper_when_created(wrapper):
+            d = wrapper.create(store)
+            d.addCallback(lambda doc: wrapper)
+            return d
+
+        d = store.list_indexes()
+        d.addCallback(try_to_get_doc_from_index)
+        d.addCallback(get_first_doc_if_any)
+        d.addCallback(wrap_existing_or_create_new)
+        return d
```
meskio: This method is long. Maybe `try_to_get_doc_from_index` and `wrap_existing_or_create_new` can become methods of the class?

src/leap/mail/adaptors/soledad.py
meskio: Ups, my bad, I didn't realize of the `None` parameter on the `kwargs.pop`. It might be still nice to define them as params with default value `None`, but not so important.

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+
+        def wrap_docs(docs):
+            return (cls(doc_id=doc.doc_id, **doc.content) for doc in docs)
+
+        d = store.get_from_index(list_index, index_value)
+        d.addCallback(wrap_docs)
+        return d
+
+    # TODO
+    # [ ] get_count() ???
```
meskio: Isn't this comment better placed up in the definition of the class with the other TODO?

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        self.fdoc = FlagsDocWrapper(doc_id=fdoc_id, **fdoc)
+        self.fdoc.set_future_doc_id(self.mdoc.fdoc)
+
+        if isinstance(hdoc, SoledadDocument):
+            hdoc_id = hdoc.doc_id
+            hdoc = hdoc.content
+        else:
+            hdoc_id = None
+        self.hdoc = HeaderDocWrapper(doc_id=hdoc_id, **hdoc)
+        self.hdoc.set_future_doc_id(self.mdoc.hdoc)
```
meskio: What about something like:

```python
def get_doc_wrapper(doc, cls):
    if isinstance(doc, SoledadDocument):
        doc_id = doc.doc_id
        doc = doc.content
    else:
        doc_id = None
    if not doc:
        doc = {}
    return cls(doc_id=doc_id, **doc)

self.mdoc = get_doc_wrapper(mdoc, MetaMsgDocWrapper)
self.fdoc = get_doc_wrapper(fdoc, MetaMsgDocWrapper)
self.fdoc.set_future_doc_id(self.mdoc.fdoc)
self.hdoc = get_doc_wrapper(hdoc, MetaMsgDocWrapper)
self.hdoc.set_future_doc_id(self.mdoc.hdoc)
```

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+
+    def __init__(self, mdoc, fdoc, hdoc, cdocs=None):
+        """
+        Need at least a metamsg-document, a flag-document and a header-document
+        to instantiate a MessageWrapper. Content-documents can be retrieved
+        lazily.
+
+        cdocs, if any, should be a dictionary in which the keys are ascending
+        integers, beginning at one, and the values are dictionaries with the
+        content of the content-docs.
```
meskio: What is the benefit of using a dictionary here instead of a list? Or is just because somewhere else is using this format already and it's easier to keep it like that.

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        leap_assert(self.mdoc.doc_id is None,
+                    "Cannot create: mdoc has a doc_id")
+        leap_assert(self.fdoc.doc_id is None,
+                    "Cannot create: fdoc has a doc_id")
+
+        # TODO check that the doc_ids in the mdoc are coherent
+        # TODO I think we need to tolerate the no hdoc.doc_id case, for when we
+        # are doing a copy to another mailbox.
+        # leap_assert(self.hdoc.doc_id is None,
+        # "Cannot create: hdoc has a doc_id")
```
meskio: I don't get this second TODO. The code is tolerating the case of no hdoc.doc_id right now, isn't it? Can this TODO and the commented code be removed? Or does the TODO need more explanation?

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        leap_assert_type(tags, tuple)
+        self.fdoc.tags = list(tags)
+
+    def set_date(self, date):
+        # XXX assert valid date format
+        self.hdoc.date = date
+
+    def get_subpart_dict(self, index):
+        """
+        :param index: index, 1-indexed
```
meskio: What is `1-indexed`?

meskio: Now I understand that `1-indexed` means that the indexes start from 1. I'm not sure if it should be more verbose, but at least I think the `index, ` is making it more confusing and should be removed.

src/leap/mail/adaptors/soledad.py
meskio: I think I understand that the point of the dictionary is to use indexes starting from 1. Not sure if I like to use a dictionary for that, but it's not a trivial change to do, I'm fine with keeping it like that.

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        :param cdocs: None, or a dictionary mapping integers (1-indexed) to
+                      dicts from where a ContentDocWrapper can be initialized.
+        :type cdocs: dict, or None
+
+        :rtype: MessageClass instance.
+        """
+        assert(MessageClass is not None)
+        return MessageClass(MessageWrapper(mdoc, fdoc, hdoc, cdocs), uid=uid)
+
+    def _get_msg_from_variable_doc_list(self, doc_list, msg_class, uid=None):
```
meskio: As this method is used by `get_msg_from_mdoc_id` I think it makes the code easier to read moving it below where is used instead of on top.

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        if len(doc_list) == 3:
+            mdoc, fdoc, hdoc = doc_list
+            cdocs = None
+        elif len(doc_list) > 3:
+            fdoc, hdoc = doc_list[:3]
```
meskio: Is mdoc missing here?

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+
+        if get_cdocs:
+            d = store.get_doc(mdoc_id)
+            d.addCallback(wrap_meta_doc)
+            d.addCallback(get_part_docs_from_mdoc_wrapper)
+        else:
+            d = get_parts_doc_from_mdoc_id()
+
+        d.addCallback(partial(self._get_msg_from_variable_doc_list,
+                              msg_class=MessageClass, uid=uid))
```
meskio: Why use `partial`? You can pass *args and **kwargs to `addCallback`.

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        return d
+
+    def create_msg(self, store, msg):
+        """
+        :param store: an instance of soledad, or anything that behaves alike
+        :type store:
```
meskio: This line could be removed or added a type to it.

src/leap/mail/adaptors/soledad.py
```diff
@@ -0,0 +1,1047 @@
...
+        :return: a Deferred that is fired when all the underlying documents
+                 have been created.
+        :rtype: defer.Deferred
+        """
+        wrapper = msg.get_wrapper()
+        return wrapper.create(store)
+
+    def update_msg(self, store, msg):
+        """
+        :param msg: a Message object.
```
meskio: Duplicated param definition.

src/leap/mail/imap/account.py
```diff
@@ -46,107 +46,56 @@ def _debugProfiling(result, cmdname, start):
...
+    def __init__(self, user_id, store, d=defer.Deferred()):
         """
-        leap_assert(soledad, "Need a soledad instance to initialize")
-        leap_assert_type(soledad, Soledad)
+        Keeps track of the mailboxes and subscriptions handled by this account.
 
-        # XXX SHOULD assert too that the name matches the user/uuid with which
-        # soledad has been initialized.
+        :param account: The name of the account (user id).
+        :type account: str
```
meskio: The param name is `user_id` instead of `account`.

src/leap/mail/imap/account.py
```diff
@@ -46,107 +46,56 @@ def _debugProfiling(result, cmdname, start):
 
 
 #######################################
-# Soledad Account
+# Soledad IMAP Account
 #######################################
 
+# XXX watchout, account needs to be ready... so we should maybe return
+# a deferred to the IMAP service when it's initialized
```
meskio: Instead of this comment we could maybe document `callWhenReady` method or put this warning in the class docstring.

src/leap/mail/imap/account.py
```diff
@@ -46,107 +46,56 @@ def _debugProfiling(result, cmdname, start):
...
-
-    @property
-    def subscriptions(self):
-        """
-        A list of the current subscriptions for this account.
-        """
-        return [doc.content[self.MBOX_KEY]
-                for doc in self._soledad.get_from_index(
-                    self.TYPE_SUBS_IDX, self.MBOX_KEY, '1')]
+    def callWhenReady(self, cb, *args, **kw):
```
meskio: This might need some documentation. I find a bit of confusing that cb's first argument is something undetermined and not `*args`. I would even prefer to change the implementation so the only params of cb are `*args` and `**kw`.

src/leap/mail/imap/account.py
```diff
@@ -46,107 +46,56 @@ def _debugProfiling(result, cmdname, start):
...
-        return doc[0] if doc else None
-
-    @property
-    def mailboxes(self):
-        """
-        A list of the current mailboxes for this account.
-        :rtype: set
-        """
-        return sorted(self.__mailboxes)
+    def _return_mailbox_from_collection(self, collection, readwrite=1):
```
meskio: As commented before, I find more readable if the private methods are placed below the ones that use them instead the other way around.

src/leap/mail/imap/account.py
```diff
@@ -182,16 +136,14 @@ def addMailbox(self, name, creation_ts=None):
                             one is provided.
         :type creation_ts: int
 
-        :returns: True if successful
+        :returns: a Deferred that will contain the document if successful.
         :rtype: bool
```
meskio: :rtype: Deferred

src/leap/mail/imap/account.py
```diff
@@ -199,44 +151,75 @@ def addMailbox(self, name, creation_ts=None):
...
+
+        def all_good(result):
+            return all(result)
+
+        if subs:
+            d1 = defer.gatherResults(subs)
+            d1.addCallback(all_good)
+            return d1
+        else:
+            return defer.succeed(False)
```
meskio: The else is not needed as subs will always contain at least `df`.

src/leap/mail/imap/account.py
```diff
@@ -199,44 +151,75 @@ def addMailbox(self, name, creation_ts=None):
...
+                return defer.succeed(False)
+            else:
+                return defer.succeed(True)
+
+        df = self.addMailbox(sep.join(paths))
+        df.addErrback(handle_collision)
+        subs.append(df)
+
+        def all_good(result):
+            return all(result)
```
meskio: The code might be more readable if all the functions are together without code intercalated on them. It took me more than 5 minutes to realize that in between the functions there was a for that was not a callback function.

src/leap/mail/imap/account.py
```diff
@@ -199,44 +151,75 @@ def addMailbox(self, name, creation_ts=None):
...
         :type pathspec: str
 
-        :return: A true value if the creation succeeds.
-        :rtype: bool
+        :return:
+            A deferred that will fire with a true value if the creation
+            succeeds.
+        :rtype: Deferred
 
         :raise MailboxException: Raised if this mailbox cannot be added.
```
meskio: There is no exception raised in this method, I'm not sure if this might be a possible failure of the Deferred.

src/leap/mail/imap/account.py
```diff
@@ -248,66 +231,92 @@ def select(self, name, readwrite=1):
...
-        name = self._parse_mailbox_name(name)
+        name = normalize_mailbox(name)
+        _mboxes = None
+
+        def check_it_exists(mailboxes):
+            global _mboxes
+            _mboxes = mailboxes
+            if name not in mailboxes:
+                err = imap4.MailboxException("No such mailbox: %r" % name)
+                return defer.fail(err)
```
meskio: raise the exception will create the same result and I think is easier to read.

src/leap/mail/imap/account.py
```diff
@@ -248,66 +231,92 @@ def select(self, name, readwrite=1):
...
+            if MessageFlags.NOSELECT_FLAG in mbox_flags:
                 # Check for hierarchically inferior mailboxes with this one
                 # as part of their root.
-                for others in self.mailboxes:
+                for others in _mboxes:
                     if others != name and others.startswith(name):
-                        raise imap4.MailboxException, (
+                        err = imap4.MailboxException(
                             "Hierarchically inferior mailboxes "
                             "exist and \\Noselect is set")
```
meskio: Again I think raise is more clear.

src/leap/mail/imap/account.py
```diff
@@ -248,66 +231,92 @@ def select(self, name, readwrite=1):
...
         # if there are no hierarchically inferior names, we will
         # delete it from our ken.
+        # XXX is this right?
         # if self._inferiorNames(name) > 1:
-        #  ??! -- can this be rite?
-        # self._index.removeMailbox(name)
+        #   self._index.removeMailbox(name)
 
+    # TODO use mail.Account.rename_mailbox
+    # TODO finish conversion to deferreds
```
meskio: I don't understand this TODOs. If I understand correctly the code is already using mail.Account.rename_mailbox and I don't see what is missing to convert to deferreds.

src/leap/mail/imap/account.py
```diff
@@ -348,54 +361,87 @@ def _inferiorNames(self, name):
...
         # XXX use wildcard query instead
-        inferiors = []
-        for infname in self.mailboxes:
-            if infname.startswith(name):
-                inferiors.append(infname)
-        return inferiors
+        def filter_inferiors(mailboxes):
+            inferiors = []
+            for infname in mailboxes:
+                if infname.startswith(name):
```
meskio: I think this should be `name.startswith(ifname)`.

src/leap/mail/imap/account.py
```diff
@@ -403,34 +449,24 @@ def unsubscribe(self, name):
...
+            return mbox.collection.set_mbox_attr("subscribed", False)
 
-        from rfc 3501:
-        returns a subset of names from the complete set
-        of all names available to the client.  Zero or more untagged LIST
-        replies are returned, containing the name attributes, hierarchy
-        delimiter, and name.
+        d = self.getMailbox(name)
+        d.addCallback(set_unsubscribed)
+        return d
```
meskio: `subscribe` and `unsubscribe` have excatly the same code except for the boolean in the set_mbox_attr. Why not create a `def _set_subscribed(self, name, subscribed):` and call it from both?

src/leap/mail/imap/mailbox.py
```diff
@@ -70,33 +73,29 @@ def do_profile_cmd(d, name):
...
-    which we instantiate and make accessible in the `messages` attribute.
+    The low-level database methods are contained in IMAPMessageCollection
+    class, which we instantiate and make accessible in the `messages`
+    attribute.
     """
     implements(
         imap4.IMailbox,
         imap4.IMailboxInfo,
-        imap4.ICloseableMailbox,
+        #imap4.ICloseableMailbox,
```
meskio: Do you keep this commented because you are planning to add support to this interface as well? In this case can you add a TODO or a XXX in front so we keep it as a task to do one day?

src/leap/mail/imap/mailbox.py
```diff
@@ -104,65 +103,31 @@ class SoledadMailbox(WithMsgFields, MBoxParser):
...
-        :param soledad: a Soledad instance.
-        :type soledad: Soledad
-
-        :param memstore: a MemoryStore instance
-        :type memstore: MemoryStore
+        :param collection: instance of IMAPMessageCollection
+        :type collection: IMAPMessageCollection
 
         :param rw: read-and-write flag for this mailbox
         :type rw: int
```
meskio: Why is rw an int instead of a bool? all the uses I find in leap.mail use it as a boolean. This is not specific to this pull and might be better to fix it in another one.

src/leap/mail/imap/mailbox.py
```diff
@@ -104,65 +103,31 @@ class SoledadMailbox(WithMsgFields, MBoxParser):
...
-
-    def __init__(self, mbox, soledad, memstore, rw=1):
+    def __init__(self, collection, rw=1):
         """
-        SoledadMailbox constructor. Needs to get passed a name, plus a
-        Soledad instance.
-
-        :param mbox: the mailbox name
-        :type mbox: str
+        SoledadMailbox constructor.
```
meskio: This class is not SoledadMailbox. It might be better to don't put a comment in this case, it don't say anything that you don't know already and it's easy to get it out of sync.

src/leap/mail/imap/mailbox.py
```diff
@@ -104,65 +103,31 @@ class SoledadMailbox(WithMsgFields, MBoxParser):
...
-
-        self.mbox = self._parse_mailbox_name(mbox)
         self.rw = rw
-
-        self._soledad = soledad
-        self._memstore = memstore
-
-        self.messages = MessageCollection(
-            mbox=mbox, soledad=self._soledad, memstore=self._memstore)
+        self.closed = False
```
meskio: What is this meant for? I don't see it used anywhere meaningfully, and it's confusing with `is_closed` and `set_closed` that are not related to it.

src/leap/mail/imap/mailbox.py
```diff
@@ -104,65 +103,31 @@ class SoledadMailbox(WithMsgFields, MBoxParser):
...
 
         self._uidvalidity = None
```
meskio: I don't see this attribute used anywhere. Can it be deleted?

src/leap/mail/imap/mailbox.py
```diff
@@ -104,65 +103,31 @@ class SoledadMailbox(WithMsgFields, MBoxParser):
...
+        self.collection = collection
 
-        # XXX careful with this get/set (it would be
-        # hitting db unconditionally, move to memstore too)
-        # Now it's returning a fixed amount of flags from mem
-        # as a workaround.
         if not self.getFlags():
```
meskio: If I understand correctly it's implementation if there is no flags `getFlags` returns `self.init_flags` so the `setFlags` will never be executed.

src/leap/mail/imap/mailbox.py
```diff
@@ -345,9 +244,10 @@ def getUID(self, message):
 
         :rtype: int
         """
-        msg = self.messages.get_msg_by_uid(message)
-        if msg is not None:
-            return msg.getUID()
+        # TODO --- return the uid if it has it!!!
+        d = self.collection.get_msg_by_uid(message)
+        d.addCallback(lambda m: m.getUID())
+        return d
```
meskio: I'm confused with this function.  As the docstring states `message` is _the message uid_, I guess this is not the same than the one returned by the function. Reading the TODO I think I understand that this function is not returning the right uid, but is returning the restul of `m.getUID()`. I'm starting to think that there is three different things named uid. Can we come up with better names for the things? or at least improve the docstring?

src/leap/mail/imap/mailbox.py
```diff
@@ -639,21 +533,41 @@ def _do_fetch(self, messages_asked, uid, d):
 
         sequence = False
         # sequence = True if uid == 0 else False
```
meskio: This comment don't match the actual implementation where sequence is always equal to False.

meskio: It's confusing to have two different things called sequence. One is this boolean, that I'm not sure what it means, and the other the sequence of filtered messages. Can we rename this boolean to something more meaningful, or remove it as it don't do anything.

src/leap/mail/imap/mailbox.py
```diff
@@ -639,21 +533,41 @@ def _do_fetch(self, messages_asked, uid, d):
...
+
+            d = defer.gatherResults(d_msg)
+            d.addCallback(_zip_msgid)
+            return d
 
         # for sequence numbers (uid = 0)
         if sequence:
             logger.debug("Getting msg by index: INEFFICIENT call!")
+            # TODO --- implement sequences in mailbox indexer
             raise NotImplementedError
```
meskio: This code will never be executed as sequence is always False.

src/leap/mail/imap/messages.py
```diff
@@ -15,188 +15,36 @@
...
-    # TODO this has to change.
-    # Should index primarily by chash, and keep a local-only
-    # UID table.
-
     implements(imap4.IMessage)
 
-    def __init__(self, soledad, uid, mbox, collection=None, container=None):
+    def __init__(self, message):
         """
         Initializes a LeapMessage.
```
meskio: A docstring with wrong information that don't say anything that we don't know.

src/leap/mail/imap/messages.py
```diff
@@ -216,62 +64,30 @@ def getFlags(self):
...
+        #:type flags: tuple of str
+        #:param mode: the mode for setting. 1 is append, -1 is remove, 0 set.
+        #:type mode: int
+        #"""
+        #leap_assert(isinstance(flags, tuple), "flags need to be a tuple")
+        # XXX
+        # return new flags
+        # map to str
+        #self.message.set_flags(flags, mode)
+        #self.collection.update_flags(self.message, flags, mode)
```
meskio: Is this waiting for collection to implement setFlags? Or can it already be deleted? I see this function is used, and will fail in `IMAPMessageCollection.set_flags`.

src/leap/mail/imap/server.py
```diff
@@ -311,21 +322,201 @@ def arg_literal(self, line):
...
+    select_DELETE = auth_DELETE
+
     # Need to override the command table after patching
     # arg_astring and arg_literal
 
+    # do_DELETE = imap4.IMAP4Server.do_DELETE
+    # do_CREATE = imap4.IMAP4Server.do_CREATE
+    # do_RENAME = imap4.IMAP4Server.do_RENAME
+    # do_SUBSCRIBE = imap4.IMAP4Server.do_SUBSCRIBE
+    # do_UNSUBSCRIBE = imap4.IMAP4Server.do_UNSUBSCRIBE
```
meskio: I'm in a quest against commented code. Is there a reason for this to stay? Can you add a comment with a TODO or XXX explaining the reason for it to stay or delete the code?

src/leap/mail/imap/service/imap.py
```diff
@@ -31,43 +30,13 @@
 
 from leap.common import events as leap_events
 from leap.common.check import leap_assert, leap_assert_type, leap_check
```
meskio: leap_assert is not used.

src/leap/mail/imap/service/imap.py
```diff
@@ -139,83 +108,40 @@ def buildProtocol(self, addr):
...
+              the factory for the protocol.
     """
-    from twisted.internet import reactor
-    # it looks like qtreactor does not honor this,
-    # but other reactors should.
-    reactor.suggestThreadPoolSize(20)
-
-    leap_assert(len(args) == 2)
-    soledad, keymanager = args
+    soledad = args[0]
```
meskio: Can we make soledad explicit on the function definition as it is needed? With this we could remove the `*args` from the definition as is not used for anything else.

src/leap/mail/imap/tests/test_imap.py
```diff
@@ -96,9 +88,10 @@ def setUp(self):
         MessageCollection interface in this particular TestCase
         """
         super(MessageCollectionTestCase, self).setUp()
```
meskio: `IMAP4HelperMixin.setUp()` returns a deferred that you are not returning here.

src/leap/mail/imap/tests/test_imap.py
```diff
@@ -96,9 +88,10 @@ def setUp(self):
...
-        memstore = MemoryStore()
-        self.messages = MessageCollection("testmbox%s" % (self.count,),
-                                          self._soledad, memstore=memstore)
+
+        # FIXME --- update initialization
+        self.messages = IMAPMessageCollection(
+            "testmbox%s" % (self.count,), self._soledad)
```
meskio: All the tests fails because this initialization wrong.

src/leap/mail/imap/tests/test_imap.py
```diff
@@ -207,23 +200,18 @@ def add_1():
         add_1().addCallback(lambda ignored: self.assertEqual(
                             mc.count(), 3))
 
-        # XXX this has to be redone to fit memstore ------------#
-        #newmsg = mc._get_empty_doc()
-        #newmsg['mailbox'] = "mailbox/foo"
-        #mc._soledad.create_doc(newmsg)
-        #self.assertEqual(mc.count(), 3)
-        #self.assertEqual(
-        #len(mc._soledad.get_from_index(mc.TYPE_IDX, "flags")), 4)
```
meskio: :+1:

src/leap/mail/imap/tests/test_imap.py
```diff
@@ -435,8 +447,12 @@ def testHierarchicalRename(self):
         """
         Try to rename hierarchical mailboxes
         """
```
meskio: This test fails for me:
```python
Traceback (most recent call last):
  File "/home/meskio/dev/leap/leap_mail/src/leap/mail/imap/tests/test_imap.py", line 474, in _cbTestHierarchicalRename
    self.assertEqual(sorted(mailboxes), sorted([s for s in expected]))
  File "/home/meskio/dev/leap/env/local/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 356, in assertEqual
    % (msg, pformat(first), pformat(second)))
twisted.trial.unittest.FailTest: not equal:
a = [u'INBOX', u'oldmbox/m1', u'oldmbox/m2']
b = ['INBOX', 'newname', 'newname/m1', 'newname/m2']
```

src/leap/mail/imap/server.py
```diff
@@ -311,21 +322,201 @@ def arg_literal(self, line):
...
+                        map(imap4.DontQuoteMe, flags),
+                        delim, name.encode('imap4-utf-7'))
+                self.sendUntaggedResponse(
+                    imap4.collapseNestedLists(resp))
+        self.sendPositiveResponse(tag, '%s completed' % (cmdName,))
+    # -------------------- end isSubscribed patch -----------
+
+    # TODO subscribe method had also to be changed to accomodate deferred
+    def do_SUBSCRIBE(self, tag, name):
+        print "DOING SUBSCRIBE"
```
meskio: **print**

There is more below.

src/leap/mail/imap/tests/test_imap.py
```diff
@@ -982,89 +1079,106 @@ def select():
...
-
-        self.assertEqual(
-            dict(msg.hdoc.content)['subject'],
-            'Message 2')
-        self.failUnless(m.closed)
+        return defer.gatherResults([d, d2])
+
+        # Okay, that was much fun indeed
+
+    # skipping close test: we just need expunge for now.
```
meskio: Something that I find a bit cleaner to skip a test is to rename it to `_testClose`. I found that done in keymanager and I like it. But anyway, not really important to fix.

src/leap/mail/imap/tests/test_imap.py
```diff
@@ -1075,47 +1189,53 @@ def expunged(results):
...
     """
     def _create_empty_mailbox(self):
-        LeapIMAPServer.theAccount.addMailbox('')
+        return self.server.theAccount.addMailbox('')
 
     def _create_one_mailbox(self):
-        LeapIMAPServer.theAccount.addMailbox('one')
+        return self.server.theAccount.addMailbox('one')
 
     def test_illegalMailboxCreate(self):
```
meskio: It fails for me:
```python
Traceback (most recent call last):
  File "/home/meskio/dev/leap/leap_mail/src/leap/mail/imap/tests/test_imap.py", line 1230, in test_illegalMailboxCreate
    d = self._create_empty_mailbox()
  File "/home/meskio/dev/leap/leap_mail/src/leap/mail/imap/tests/test_imap.py", line 1217, in _create_empty_mailbox
    return self.server.theAccount.addMailbox('')
  File "/home/meskio/dev/leap/leap_mail/src/leap/mail/imap/account.py", line 146, in addMailbox
    leap_assert(name, "Need a mailbox name to create a mailbox")
  File "/home/meskio/dev/leap/leap_pycommon/src/leap/common/check.py", line 48, in leap_assert
    assert condition, message
exceptions.AssertionError: Need a mailbox name to create a mailbox
```

meskio:
Missing leap/mail/imap/tests/test_imap_store_fetch.py to be updated, I guess will come in another pull-req.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+def _write_and_rewind(payload):
+    fd = StringIO.StringIO()
+    fd.write(payload)
+    fd.seek(0)
+    return fd
+
+
+class MessagePart(object):
+
+    # TODO pass cdocs in init
```
meskio: I don't get it. There is a parameter cdocs in the init. What is missing to do here?

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+
+    def __init__(self, part_map, cdocs={}):
```
meskio: It might be useful to document what is part_map.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+
+    def get_headers(self):
+        """
+        Get the raw headers document.
+        """
+        return [tuple(item) for item in self._wrapper.hdoc.headers]
+
+    def get_body_file(self, store):
+        """
+        """
```
meskio: Quotes not needed if there is no docstring.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+        # TODO --- review this count shit. I think it's better to patch server
+        # to accept deferreds.
+        # TODO need to initialize count here because imap server does not
+        # expect a defered for the count. caller should return the deferred for
+        # prime_count (ie, initialize) when returning the collection
+        # TODO should increment and decrement when adding/deleting.
+        # TODO recent count should also be static.
+
+        if not count:
+            count = 0
```
meskio: Why not use `count=0` in the definition of the method instead of doing this if? I don't know if is a pythonism, but I don't like much that things in leap code often have a default value to `None` instead to a default value that has any meaning for the variable.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+        Return True if this collection represents a Mailbox.
+        :rtype: bool
+        """
+        return bool(self.mbox_wrapper)
+
+    @property
+    def mbox_name(self):
+        wrapper = getattr(self, "mbox_wrapper", None)
+        if not wrapper:
+            return None
```
meskio: Don't need to do the `getattr` you can directly do `if self.mbox_wrapper is None:` as self.mbox_wrapper is always initialized by the constructor.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+
+    @property
+    def mbox_uuid(self):
+        wrapper = getattr(self, "mbox_wrapper", None)
+        if not wrapper:
+            return None
+        return wrapper.uuid
+
+    def get_mbox_attr(self, attr):
+        return getattr(self.mbox_wrapper, attr)
```
meskio: What if mbox_wrapper is `None`? Can we add a TODO here and in the subsequent methods that assume it exist?

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+
+    def get_message_by_content_hash(self, chash, get_cdocs=False):
+        """
+        Retrieve a message by its content hash.
+        :rtype: Deferred
+        """
+
+        if not self.is_mailbox_collection():
+            # instead of getting the metamsg by chash, query by (meta) index
+            # or use the internal collection of pointers-to-docs.
```
meskio: Is this comment a TODO or is something that comes from a copy-paste? Can you put the TODO explicit if this is the case?

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+            return self.adaptor.get_msg_from_mdoc_id(
+                self.messageklass, self.store,
+                doc_id, uid=uid, get_cdocs=get_cdocs)
+
+        d = self.mbox_indexer.get_doc_id_from_uid(self.mbox_uuid, uid)
+        d.addCallback(get_msg_from_mdoc_id)
+        return d
+
+    # TODO deprecate ??? ---
+    def _prime_count(self):
```
meskio: I don't see it used anywhere, can we delete this method?

meskio: I see it now mentioned in a TODO in the init, up to you if you think that will be used by the init we can keep it.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+        """
+        Copy the message to another collection. (it only makes sense for
+        mailbox collections)
+        """
+        if not self.is_mailbox_collection():
+            raise NotImplementedError()
+
+        def insert_copied_mdoc_id(wrapper):
+            return self.mbox_indexer.insert_doc(
+                newmailbox_uuid, wrapper.mdoc.doc_id)
```
meskio: Where is this `newmailbox_uuid` coming from? Might it be a typo and you meant `newmailbox`?

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+                d = self.get_message_by_uid(uid)
+                d.addCallback(lambda msg: msg.delete())
+                deferreds.append(d)
+            return defer.gatherResults(deferreds)
+
+        d = self.all_uid_iter()
+        d.addCallback(del_all_uid)
+        return d
+
+    def _update_flags_or_tags(self, old, new, mode):
```
meskio: Again my OCD on clean code. Can we move this method below the ones that used it?

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+class Account(object):
+    """
+    Account is the top level abstraction to access collections of messages
+    associated with a LEAP Mail Account.
+
+    It primarily handles creation and access of Mailboxes, which will be the
+    basic collection handled by traditional MUAs, but it can also handle other
+    types of Collections (tag based, for instance).
+
+    leap.mail.imap.SoledadBackedAccount partially proxies methods in this
```
meskio: It's name now is not `SoledadBackendAccount` but `IMAPAccount`.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+    class.
+    """
+
+    # Adaptor is passed to the returned MessageCollections, so if you want to
+    # use a different adaptor this is the place to change it, by subclassing
+    # the Account class.
+
+    adaptor_class = SoledadMailAdaptor
+    store = None
```
meskio: Is there any need for the `store = None` as init will rewrite it? The only case I can come up is for classmethods, but there is none in this class.

src/leap/mail/mail.py
```diff
@@ -0,0 +1,697 @@
...
+                self._ready_cb()
+
+        d = self.adaptor.initialize_store(self.store)
+        d.addCallback(lambda _: self.list_all_mailbox_names())
+        d.addCallback(add_mailbox_if_none)
+        d.addCallback(finish_initialization)
+        return d
+
+    def callWhenReady(self, cb, *args, **kw):
+        self.deferred_initialization.addCallback(cb, *args, **kw)
```
meskio: As mentioned above, I'm not sure if I would prefer to do something like:
```
self.deferred_initialization.addCallback(lambda _: cb(*args, **kw))
```
If we keep it like that I think we should document that `cb` will get a meaningless first parameter.

src/leap/mail/mailbox_indexer.py
```diff
@@ -0,0 +1,333 @@
...
+        :param mailbox: the mailbox identifier.
+        :type mailbox: str
+        :rtype: Deferred
+        """
+        check_good_uuid(mailbox_id)
+        sql = ("CREATE TABLE if not exists {preffix}{name}( "
+               "uid  INTEGER PRIMARY KEY AUTOINCREMENT, "
+               "hash TEXT UNIQUE NOT NULL)".format(
+                   preffix=self.table_preffix, name=sanitize(mailbox_id)))
+        return self._query(sql)
```
meskio: I see that everywhere you are checking that the parameters are valid to avoid sql-injections, but just as extra precaution won't be better to pass the parameters to sqlite instead of format them in the string?

src/leap/mail/mailbox_indexer.py
```diff
@@ -0,0 +1,333 @@
...
+        :type newmailbox: str
+        :rtype: Deferred
+        """
+        assert oldmailbox
+        assert newmailbox
+        assert oldmailbox != newmailbox
+        sql = ("ALTER TABLE {preffix}{old} "
+               "RENAME TO {preffix}{new}".format(
+                   preffix=self.table_preffix,
+                   old=sanitize(oldmailbox), new=sanitize(newmailbox)))
```
meskio: Here `oldmailbox` and `newmailbox` are not checked to be valid UIDs, that might lead to a sql-injection. Not sure how much we care, the only attacker here will be your MUA.

src/leap/mail/mailbox_indexer.py
```diff
@@ -0,0 +1,333 @@
...
+    def insert_doc(self, mailbox_id, doc_id):
+        """
+        Insert the doc_id for a MetaMsg in the UID table for a given mailbox.
+
+        The doc_id must be in the format:
+
+            M+<mailbox>+<content-hash-of-the-message>
+
+        :param mailbox: the mailbox name
+        :type mailbox: str
```
meskio: The param is called `mailbox_id` and I have the feeling that is not the name but the uid of it.

src/leap/mail/mailbox_indexer.py
meskio: I see you pass the parameters to sqlcipher in some cases. When you don't pass them, is it because you can not pass them in this situations?

src/leap/mail/smtp/tests/test_gateway.py
```diff
@@ -23,6 +23,7 @@
 import re
```
meskio: All the smtp tests fail, but this might be my fault as I think all the modifications here are done by me.

src/leap/mail/tests/test_mail.py
```diff
@@ -0,0 +1,378 @@
+# -*- coding: utf-8 -*-
```
meskio: Many tests here fail, but I guess we can merge the pull-req and take care of the failing tests afterwards.

src/leap/mail/walk.py
```diff
@@ -85,6 +85,35 @@
...
+              shall be known in this contract as the first part of the party of
+              the first part, shall be known in this contract - look, why
+              should we quarrel about a thing like this, we'll take it right
+              out, eh?
+
+Chico Marx: Yes, it's too long anyhow. Now what have we got left?
+Groucho Marx: Well I've got about a foot and a half. Now what's the matter?
+
+Chico Marx: I don't like the second party either.
+"""
```
meskio: :smile:

meskio:
Done, I'm going to drink some mezcales and lay down.

src/leap/mail/adaptors/soledad.py
kalikaneko: fail, fixed

src/leap/mail/adaptors/soledad.py
kalikaneko: done

src/leap/mail/adaptors/soledad.py
kalikaneko: done. I think this is a tick of mine (in part for past reviews), but in reality I don't see any benefits  for using the property boilerplate in this case...

src/leap/mail/adaptors/soledad.py
kalikaneko: hmmm... I see your point, but in this case what failures are you thinking about? they will be the generic soledad/u1db errors methings, not sure if it's useful repeating them everywhere where we perform soledad operations...

src/leap/mail/adaptors/soledad.py
kalikaneko: agree. left TODO to split it when going through tests.

src/leap/mail/adaptors/soledad.py
kalikaneko: indeed.

src/leap/mail/adaptors/soledad.py
kalikaneko: cleaner, thanks

src/leap/mail/adaptors/soledad.py
kalikaneko: Hmm I think I carry this from the part_map thing in the walk module.
I don't really know what would be more efficient for lookups (I think dict). I'll think about it.

src/leap/mail/adaptors/soledad.py
kalikaneko: removed both

src/leap/mail/adaptors/soledad.py
kalikaneko: true, that seems a riddle

src/leap/mail/adaptors/soledad.py
kalikaneko: agree, moved

src/leap/mail/adaptors/soledad.py
kalikaneko: yes!

src/leap/mail/adaptors/soledad.py
kalikaneko: hum! it seems I had a reason for this thing. The setter raises an Exception because it's not defined in the model. So leaving it with the set_future_doc_id

src/leap/mail/adaptors/soledad.py
kalikaneko: true! this must be something that evolved from a different use, or maybe a lapsus.

src/leap/mail/adaptors/soledad.py
kalikaneko: removed.

src/leap/mail/adaptors/soledad.py
kalikaneko: deleted

src/leap/mail/imap/account.py
kalikaneko: fixed

src/leap/mail/imap/account.py
kalikaneko: removed comment and improved the class docstring.

src/leap/mail/imap/account.py
kalikaneko: you're right. left a note to change and fix tests accordingly.

src/leap/mail/imap/account.py
kalikaneko: moved

src/leap/mail/imap/account.py
kalikaneko: fixed

src/leap/mail/imap/account.py
kalikaneko: you're right, bad copypasta from the non-deferred case.

src/leap/mail/imap/account.py
kalikaneko: :( reordered

src/leap/mail/imap/account.py
kalikaneko: right, outdated docstring. the error is now in the form of a deferred failure

src/leap/mail/imap/account.py
kalikaneko: changed, this suffered from progressive conversion to deferreds, but now it's ok

src/leap/mail/imap/account.py
kalikaneko: yep

src/leap/mail/imap/account.py
kalikaneko: TODO remove this TODO when no longer needed

src/leap/mail/imap/account.py
kalikaneko: no, it's correct: Foo/bar y Foo/baaz empiezan con Foo y los queremos todos.

src/leap/mail/imap/account.py
kalikaneko: left TODO. I also realized that unsubscribe should raise an exception if attempted to unsubscribe from non-subscribed mailbox.

src/leap/mail/imap/mailbox.py
kalikaneko: I recently discovered that I had origininally implemented this interface but it might not be needed at all, since it seems to be optional and I don't think we need any extra cleanup other than what's done at expunge. But I want to be sure i'm not missing something, I'll leave a comment for now.

src/leap/mail/imap/mailbox.py
kalikaneko: twisted server passes us an int, I decided to cast it to bool for our use.

src/leap/mail/imap/mailbox.py
kalikaneko: agree

src/leap/mail/imap/mailbox.py
kalikaneko: Hmm... I should document this. It's used in a rather hacky way, being set when we stop the service (currently, from the imap controller in the client backend!)
```
self.imap_factory.theAccount.closed = True
```
It's used as a way of deauthenticating the clients, to bypass a privacy breach.
I think I should make it more explicit, something like end_session() or something...

src/leap/mail/imap/messages.py
kalikaneko: deleted

src/leap/mail/imap/mailbox.py
kalikaneko: I've updated this today. The function now is meaningless (but expected).
For now i'm just returning the UID (there is only one such thing), I'll deal with relative sequence numbers in other moment (TB doesn't seem to use them).

src/leap/mail/imap/mailbox.py
kalikaneko: see above (and scream)

kalikaneko: here I left the review changes today ---

src/leap/mail/imap/mailbox.py
kalikaneko: Hum.. I confused the use of "closed" on account and on mailbox.
I've cleaned up the use of the account session termination, and removed several cruft in mailbox. I think I can remove all the close attributes since I don't implement that interface anymore, but I'd like to check that carefully.

src/leap/mail/imap/mailbox.py
kalikaneko: err... yes, it's not used anywhere in mailbox. deleted

src/leap/mail/imap/mailbox.py
kalikaneko: true. removed.

src/leap/mail/adaptors/soledad.py
meskio: I'm fine having a setter, I was wondering about using the property for the getter but not for the setter. Actually as there is no logic involved here it will be the same to have the attribute public. But I'm fine to keep it like that.

src/leap/mail/imap/mailbox.py
kalikaneko: it's true, this is confusing as hell. I've renamed the bool to is_sequence (which is not supported right now, will need changes in mailbox_indexer), and I called the other "sequences" message_ranges

src/leap/mail/imap/mailbox.py
kalikaneko: fixed, left the exception path executable

src/leap/mail/adaptors/soledad.py
meskio: I have the feeling that we don't check any kind of failures, and that bitmask will crash frequently by stupid failures, or that some remote attackers could provoke crashes in bitmask. But it's pretty hard to check failures if you don't know what can fail, I have no idea myself of the kinds of failures that soledad can return, AFAIK the are neither documented there.

I think we should document the failures that the public APIs can return. But that is something that we can try to improve in following iterations.

src/leap/mail/imap/account.py
meskio: Ahh, I understood the logic of the function wrongly. Now I get it.

src/leap/mail/imap/messages.py
kalikaneko: true again, thanks for reading XD
I have attempted to be more informative now.

src/leap/mail/imap/server.py
kalikaneko: I'd call this case special (You see how I like to do that!) because it's more or less a patch that could be cleaned up and sent usptream, so I'ts useful for me to keep the list of the methods in the twisted imap server affected by the patch. Those are commented because, for some reason (that I don't remember), i've overriden the methods in our implementation too. 

I've marked those blocks a bit better.

src/leap/mail/imap/service/imap.py
kalikaneko: I assert that is true, fixed

src/leap/mail/imap/service/imap.py
kalikaneko: done, i've called it "store". Although the only store in a while will be a Soledad instance, I'm more confortable having it as a generic type in case we can to change something (and that's the idea behind the adaptor).

src/leap/mail/imap/tests/test_imap.py
kalikaneko: true. I haven't payed attention to this testcase, it probably have to be redone or moved to some other module. added TODO

src/leap/mail/imap/tests/test_imap.py
kalikaneko: yes, thanks for noticing. rename's broken, I expect to fix the testsuites while we're doing some generic QA on the internal test branch.

src/leap/mail/imap/server.py
kalikaneko: removed

src/leap/mail/imap/tests/test_imap.py
kalikaneko: I removed the test, since I removed the implementor of the interface.

src/leap/mail/imap/tests/test_imap.py
kalikaneko: thanks. this testsuite is also unifinished, test needs some rework.

src/leap/mail/mail.py
kalikaneko: old comment is old. thanks, removed.

src/leap/mail/mail.py
kalikaneko: now there is :)

src/leap/mail/mail.py
kalikaneko: I kind of did it now, although it should be more verbose about the expected format. This is an obscure corner, which is rather untested yet by the way.

src/leap/mail/mail.py
kalikaneko: that was part of an idea I retracted of, but point taken, I'll cleanup whenever I can.

src/leap/mail/mail.py
kalikaneko: it is now, changed.

src/leap/mail/mail.py
kalikaneko: good point. I've changed it to raise an exception if the method is used with a collection that is not a mailbox collection.
This makes me think that this class is badly designed, and that the MailboxCollection could be better implemented by subclassing... what do you think?

src/leap/mail/mail.py
kalikaneko: it's a TODO, added.

src/leap/mail/mail.py
kalikaneko: removed. this is part of a series of cases in which I've got things that *are* deferreds, but the server implementation expects inmediate values. I've considered patching the server to use maybeDeferred, but maybe it's easier to wait for intializations (in our __init__ antipattern, or in a factory function as exarkun recommends in his post) and then just handle the values.

I think I'll tackle the several count_ methods right after we merge this onto async-api, since it's needed for the notifications to the MUA to work properly.

src/leap/mail/mail.py
kalikaneko: it doesn't exist, I postponed the implementation of the copy interface.
I added a little TODO with a stub of the function that needs to be created.

src/leap/mail/mail.py
kalikaneko: done, it's prettier this way

src/leap/mail/mail.py
kalikaneko: done, thanks

src/leap/mail/mail.py
kalikaneko: No, removed.
At first I attempted to let the factory to set the attribute after creating the instance,
but for some reason I ended up making it mandatory in the initialziation.

src/leap/mail/mail.py
kalikaneko: yep, left a note with this, and added comment to docstring.

src/leap/mail/mailbox_indexer.py
kalikaneko: afaik you cannot pass the table name as a parameter

src/leap/mail/mailbox_indexer.py
kalikaneko: Hmm ... I'd swear I had deleted this method.
This doesn't make sense now, because mailboxes uuid won't change, even if we rename them.
Removed.

src/leap/mail/mailbox_indexer.py
kalikaneko: actually, it's an UUID, so I renamed it to mailbox_uuid for clarity.

src/leap/mail/smtp/tests/test_gateway.py
kalikaneko: Cool, let's merge and attack the tests. I haven't tested smtp beyond fixing its initialization yet.

src/leap/mail/tests/test_mail.py
kalikaneko: :+1:

kalikaneko:
done with fixes! thanks for the extensive reading and review :neckbeard:

kalikaneko:
did a force update. I've squashed and reordered some redundant commits in the branch, and rephrase some of the commit messages. most of the fixes I did are in the last commit for ease of review.
If you are ok with changes, I'd say we merge this into feature/async-api so we can start the 0.9.0a1 bundle dance :)
(btw, let's do a merge-commit on this one to remind us that the intermediate state surely it's not working).

meskio:
Merged into the `feature/async-api` branch.

PR #159, opened 2015-01-03
Port `enum` to `enum34`

meskio:


kalikaneko:
merged, but had to rebase against develop so it's not automatically closed.

PR #165, opened 2015-01-28
Bug/copy interface

kalikaneko:
implements the copiable interface,
and adds some performance tweaks for both copy and append operations.

### DIFF
--- a/src/leap/mail/adaptors/soledad.py
+++ b/src/leap/mail/adaptors/soledad.py
@@ -18,12 +18,14 @@ Soledadad MailAdaptor module.
 """
 import logging
 import re
+
 from collections import defaultdict
 from email import message_from_string
 
 from pycryptopp.hash import sha256
 from twisted.internet import defer
 from zope.interface import implements
+import u1db
 
 from leap.common.check import leap_assert, leap_assert_type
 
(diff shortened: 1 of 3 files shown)


src/leap/mail/adaptors/soledad.py
```diff
@@ -483,17 +529,24 @@ def create(self, store):
...
+                if cdoc.doc_id is not None:
+                    # we could be just linking to an existing
+                    # content-doc.
+                    continue
+                d.append(cdoc.create(store))
+
+        if notify_just_mdoc:
+            return mdoc_created
+        else:
+            return defer.gatherResults(d)
```
meskio: AFAIK if you don't add errbacks and the deferred fails the program will crash when the garbage collector tries to free it. For the case of notify_just_mdoc we might need to add a dummy errback.

kalikaneko: Hmm. I see your point, but I'm afraid that if I catch the exceptions in here I'll be avoiding the failure to propagate to upper layers, and thus invalidating more meaningfull errBacks (at the level of IMAP server methods, for instance: there are always methods adding callbacks and errbacks to these soledad deferreds, and it makes more sense to catch errors at that level - notifying to the MUA that an append has failed, for instance).

I think I want to make some real tests to see how this would behave before taking action.

meskio: Ok, I'm fine with it then. I'll merge it.

PR #166, opened 2015-01-28
Add public key as attachment

meskio:


elijh:
I wonder if there is anything we can do to make the filename be less english specific.

"email-encryption-key-for-user@domain.org" is good in english, but not so good otherwise.

elijh:
maybe the file should have .asc file suffix. This is what enigmail does when it attaches your public key.

elijh:
how about?: "user@domain.org-email-key.asc"

"email" is probably one word that scans in a lot of languages.

kalikaneko:
> how about?: "user@domain.org-email-key.asc"
:+1:

meskio:
Sounds good to me the new proposed name. Changed.

PR #169, opened 2015-02-18
Feature/6742 notify inbox insertions

meskio:
This remove all the dependencies of imap from the incoming. The event's are still called IMAP_*.

This has to be merged together with: https://github.com/leapcode/bitmask_client/pull/823

### DIFF
--- a/src/leap/mail/imap/mailbox.py
+++ b/src/leap/mail/imap/mailbox.py
@@ -19,13 +19,9 @@ IMAP Mailbox.
 """
 import re
 import logging
-import StringIO
-import cStringIO
-import time
 import os
 
 from collections import defaultdict
-from email.utils import formatdate
 
 from twisted.internet import defer
 from twisted.internet import reactor
@@ -36,7 +32,7 @@ from zope.interface import implements
 
 from leap.common import events as leap_events
 from leap.common.events.events_pb2 import IMAP_UNREAD_MAIL
-from leap.common.check import leap_assert, leap_assert_type
+from leap.common.check import leap_assert
 from leap.mail.constants import INBOX_NAME, MessageFlags
 from leap.mail.imap.messages import IMAPMessage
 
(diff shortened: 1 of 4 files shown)


meskio:
@dudadornelles this pull-req implements the inbox insertion callback that we had talk about. I hope it looks good to you.

src/leap/mail/imap/mailbox.py
```diff
@@ -326,41 +322,7 @@ def addMessage(self, message, flags, date=None):
...
-        # This will not catch any errors when the inserts of the other parts
-        # fail, but on the other hand allows us to return very quickly, which
-        # seems a good compromise given that we have to serialize the appends.
-        # A better solution will probably involve implementing MULTIAPPEND
-        # or patching imap server to support pipelining.
-
-        d = self.collection.add_msg(message, flags=flags, date=date,
-                                    notify_just_mdoc=True)
-
-        # XXX signal to UI? --- should do it only if INBOX...
```
kalikaneko: can you leave this XXX line there? it's something that still needs to be addressed
also, beware that you're removing the notify_just_mdoc here.

src/leap/mail/imap/mailbox.py
```diff
@@ -326,41 +322,7 @@ def addMessage(self, message, flags, date=None):
...
-        # XXX we could treat the message as an IMessage from here
-        leap_assert_type(message, basestring)
-
-        if flags is None:
-            flags = tuple()
-        else:
-            flags = tuple(str(flag) for flag in flags)
-
-        if date is None:
-            date = formatdate(time.time())
```
kalikaneko: why remove this?

src/leap/mail/imap/mailbox.py
```diff
@@ -326,41 +322,7 @@ def addMessage(self, message, flags, date=None):
         :return: a deferred that will be triggered with the UID of the added
                  message.
         """
-        # TODO should raise ReadOnlyMailbox if not rw.
-        # TODO have a look at the cases for internal date in the rfc
-        if isinstance(message, (cStringIO.OutputType, StringIO.StringIO)):
-            message = message.getvalue()
```
kalikaneko: removed?

src/leap/mail/imap/mailbox.py
kalikaneko: uuh, sorry, I see all this block has been moved... :/

meskio: Actually the XXX I removed, AFAIK the UI is signaled from the IncomingMail service:
https://github.com/meskio/leap_mail/blob/feature/6742_notify_inbox_insertions/src/leap/mail/incoming/service.py#L727

Do you think that imap or mail should signal it again, or with different signal?

meskio:
True, I see that there is one major difference here with twisted-callbacks, that here they will be called multiple times instead of once. For the rest the concept is basically the same than twisted-callbacks.

I guess we can go for add_listener. I'll fix it.

PR #171, opened 2015-02-19
Bug/fix bad attachment rendering

kalikaneko:


src/leap/mail/imap/mailbox.py
```diff
@@ -462,7 +462,8 @@ def fetch(self, messages_asked, uid):
         # Change this to the flag that twisted expects when we
         # switch to content-hash based index + local UID table.
 
-        is_sequence = True if uid == 0 else False
+        #is_sequence = True if uid == 0 else False
+        is_sequence = False # lie to me so that my client works
```
meskio: Whot? Can't we fix that? Or at least add a XXX comment explaining it?

src/leap/mail/imap/service/imap-server.tac
```diff
@@ -131,6 +110,7 @@ tempdir = "/tmp/"
 
 # Ad-hoc soledad/keymanager initialization.
 
+print "[~] user:", userid
```
meskio: I'm not sure how tac files work, do we really want a print here?

meskio: Ups, sorry, I see there is more prints and is normal. Ignore.

src/leap/mail/imap/mailbox.py
kalikaneko: I can add a better explanation, sure, this is a debug-hack. The client I'm referring to here is the twisted.mail.imap.client, that sends sequence numbers instead of absolute uids, and adding full support for that is going to take a while.

src/leap/mail/imap/service/imap-server.tac
```diff
@@ -144,6 +124,8 @@ km_kwargs = {
 }
 keymanager = KeyManager(*km_args, **km_kwargs)
 
+# XXX Do we need to wait until keymanager is properly initialized?
+
```
meskio: There is no need to wait for keymanager, as soon as the constructor returns you can use the object without problems, the only exception is the method `get_all_keys` that might fail if the initialization is not done, but this method is not used here.

src/leap/mail/imap/service/imap.py
```diff
@@ -136,7 +136,9 @@ def run_service(store, **kwargs):
     :returns: the port as returned by the reactor when starts listening, and
               the factory for the protocol.
     """
-    leap_assert_type(store, Soledad)
+    leap_check(store, "store cannot be None")
+    # XXX this can also be a ProxiedObject, FIXME
+    # leap_assert_type(store, Soledad)
```
meskio: To fit with my OCR I'll prefer the comment on top of the `leap_check`.

kalikaneko: it is above the line that it refers to

src/leap/mail/imap/service/imap-server.tac
kalikaneko: cool, will remove this then, thanks

src/leap/mail/imap/mailbox.py
kalikaneko: fixed

docs/conf.py
```diff
@@ -59,7 +59,7 @@
 # The short X.Y version.
 version = '0.4.0alpha1'
```
ivanalejandro0: does this need to change along with `release`?

ivanalejandro0:
rebased for a clean merge and merged on `develop`.

PR #172, opened 2015-02-20
Feature/send bye before disconnect

kalikaneko:


kalikaneko:
this will need rebase b4 merging, since it include the commits from pull/171

PR #173, opened 2015-02-25
Bug/inbox notifications to ui

kalikaneko:


src/leap/mail/imap/mailbox.py
```diff
@@ -20,8 +20,11 @@
 import re
 import logging
 import os
+import cStringIO
+import time
```
meskio: You can remove the time import in line 57 now.

src/leap/mail/imap/mailbox.py
```diff
@@ -322,7 +324,34 @@ def addMessage(self, message, flags, date=None):
         :return: a deferred that will be triggered with the UID of the added
                  message.
         """
-        return self.collection.add_raw_message(message, flags, date)
+        # TODO should raise ReadOnlyMailbox if not rw.
+        # TODO have a look at the cases for internal date in the rfc
+        # XXX we could treat the message as an IMessage from here
+
+        if isinstance(message, (cStringIO.OutputType, StringIO.StringIO)):
```
meskio: Can it be missing the `import StringIO`? I only see the cStringIO import. Or could you use cStringIO instead of StringIO?

src/leap/mail/incoming/service.py
```diff
@@ -729,7 +729,7 @@ def signal_deleted(doc_id):
             d.addCallback(signal_deleted)
             return d
 
-        d = self._inbox.add_raw_message(data, (self.RECENT_FLAG,))
+        d = self._inbox.add_msg(data, (self.RECENT_FLAG,))
```
meskio: Inbox here is not a imap MailBox, but a MessageCollection.

meskio: Ups, I see now that you have modified `add_msg`. I keep reading before complaining more.

meskio: `add_msg` now does not set the date as `add_raw_message` was doing. We may care more about the date that comes in the message header, but then in `add_msg` we are setting the date as "" removing the date from the header.

src/leap/mail/mail.py
```diff
@@ -556,57 +555,29 @@ def insert_mdoc_id(_, wrapper):
...
+    def cb_signal_unread_to_ui(self, result):
         """
-        Adds a message to this collection.
-
-        :param message: the raw message
-        :type message: str
-
-        :param flags: flag list
-        :type flags: list of str
+        Sends unread event to ui.
```
meskio: Can you document that the event is only sent if is inbox? I'm not sure why we signal only in this case, but I find confusing that the doc and the implementation states different things.

kalikaneko: Sure. The problem is that, in the client side, there's a message that says "N Unread messages in your Inbox"

src/leap/mail/imap/mailbox.py
kalikaneko: done, thanks

src/leap/mail/imap/mailbox.py
kalikaneko: it was missing indeed, fucking code dance

src/leap/mail/incoming/service.py
kalikaneko: right, you're right.
the internal date is not the date in the headers, according to specs this has different meanings.
I've clarified the docstring, and also I'm adding the insertion date from the incoming service now.

kalikaneko:
updated pullreq according to discussion

PR #174, opened 2015-02-27
Bug/fix mail tests 2

kalikaneko:
unfinished yet, I'm going to continue fixing some tests more, but this is more correct than develop.


kalikaneko:
I think this can be already merged, a couple of tests are randomly failing but overall they are improved compared to develop

PR #175, opened 2015-03-05
Feature/keep mapping of collections

kalikaneko:


src/leap/mail/mail.py
```diff
@@ -837,10 +854,16 @@ def get_collection_by_mailbox(self, name):
         """
         :rtype: MessageCollection
         """
+        collection = self._collection_mapping.get(name, None)
+        if collection:
+            return collection
```
andrejb: So, this method is actually returning either a MessagecCollection or a Deferred that will fire with a MessageCollection. I'm not sure if this is indeed the expected behaviour and documentation, but if not maybe we could always return a Deferred, maybe like this:

```python
    if collection:
        return defer.Deferred().addCallback(lambda _: collection)
```

Or something like this.

src/leap/mail/mail.py
```diff
@@ -560,6 +562,15 @@ def add_msg(self, raw_msg, flags=tuple(), tags=tuple(), date="",
 
         def insert_mdoc_id(_, wrapper):
             doc_id = wrapper.mdoc.doc_id
+            if not doc_id:
+                # --- BUG -----------------------------------------
+                # XXX why from time to time mdoc doesn't have doc_id
+                # here???
```
andrejb: Do you think this might be related to: https://leap.se/code/issues/6769
?

src/leap/mail/mail.py
kalikaneko: Yes, you're right.
In case you wonder how this could possibly work, it's because it was called in the middle of a deferred chain, so the result was the same.
i'll change this by a defer.Deferred(result) so that the return type is consistent.

src/leap/mail/mail.py
kalikaneko: I'm not 100% sure, but I consider it a possibility.

kalikaneko:
Fixed after review, ready to merge.

PR #176, opened 2015-03-23
Bug/several mime fixes

kalikaneko:
each commit has its own description, it will make more sense to see commit changes separately since here there are several not-too related fixes.

### DIFF
--- a/src/leap/mail/adaptors/soledad.py
+++ b/src/leap/mail/adaptors/soledad.py
@@ -1114,6 +1114,7 @@ def _split_into_parts(raw):
 
     msg, parts, chash, multi = _parse_msg(raw)
     size = len(msg.as_string())
+
     body_phash = walk.get_body_phash(msg)
 
     parts_map = walk.walk_msg_tree(parts, body_phash=body_phash)
@@ -1161,16 +1162,13 @@ def _build_headers_doc(msg, chash, body_phash, parts_map):
 
     It takes into account possibly repeated headers.
     """
-    headers = msg.items()
-
-    # TODO move this manipulation to IMAP
-    #headers = defaultdict(list)
-    #for k, v in msg.items():
-        #headers[k].append(v)
-    ## "fix" for repeated headers.
-    #for k, v in headers.items():
-        #newline = "\n%s: " % (k,)
-        #headers[k] = newline.join(v)
+    headers = defaultdict(list)
+    for k, v in msg.items():
+        headers[k].append(v)
+    # "fix" for repeated headers (as in "Received:"
+    for k, v in headers.items():
+        newline = "\n%s: " % (k.lower(),)
+        headers[k] = newline.join(v)
 
     lower_headers = lowerdict(dict(headers))
     msgid = first(_MSGID_RE.findall(
--- a/src/leap/mail/adaptors/soledad.py
+++ b/src/leap/mail/adaptors/soledad.py
@@ -491,7 +491,7 @@ class MessageWrapper(object):
         for doc_id, cdoc in zip(self.mdoc.cdocs, self.cdocs.values()):
             cdoc.set_future_doc_id(doc_id)
 
-    def create(self, store, notify_just_mdoc=False):
+    def create(self, store, notify_just_mdoc=False, pending_inserts_dict=None):
         """
         Create all the parts for this message in the store.
 
@@ -503,7 +503,7 @@ class MessageWrapper(object):
             Be warned that in that case there will be no record of failures
             when creating the other part-documents.
 
-            Other-wise, this method will return a deferred that will wait for
+            Otherwise, this method will return a deferred that will wait for
             the creation of all the part documents.
 
             Setting this flag to True is mostly a convenient workaround for the
@@ -513,6 +513,9 @@ class MessageWrapper(object):
             times will be enough to have all the queued insert operations
             finished.
         :type notify_just_mdoc: bool
+        :param pending_inserts_dict:
+            a dictionary with the pending inserts ids.
+        :type pending_inserts_dict: dict
 
         :return: a deferred whose callback will be called when either all the
                  part documents have been written, or just the metamsg-doc,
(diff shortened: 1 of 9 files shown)


kalikaneko:
e139fb9 was supposed to go right after   57ec516. I see that way in my local history.
However, I believe GH is displaying it in other ordering because I added (and rebased) that commit in a force-push.
@meskio can you check if you see the right ordering locally?

meskio:
Yes, I see the commits in the desired order, I don't know what happens in the view here.

There is a regression in the tests `leap.mail.tests.test_mail.MessageTestCase.test_get_headers` is failing and `leap.mail.imap.tests.test_imap.LEAPIMAP4ServerTestCase.testFullAppend` that was already failing now gives me two errors (before was giving only one). I didn't have a look yet to the code.

kalikaneko:
oops, sorry with the puzzle to find what was missing in the PR I forgot to review the tests. having a look.

src/leap/mail/imap/mailbox.py
```diff
@@ -500,9 +500,9 @@ def fetch(self, messages_asked, uid):
         is_sequence = True if uid == 0 else False
 
         # XXX DEBUG ---  if you attempt to use the `getmail` utility under
-        # imap/tests, it will choke until we implement sequence numbers. This
-        # is an easy hack meanwhile.
-        # is_sequence = False
+        # imap/tests, or muas like mutt, it will choke until we implement
+        # sequence numbers. This is an easy hack meanwhile.
+        is_sequence = False
```
meskio: This is ugly. Do we really need to hardcode `is_sequence` to False? In this case, can we at least remove (or comment) the previous set of `is_sequence`?

I see is a temporary fix and I will merge it like that, but I'm still not really happy to include nasty workarounds that might pile up in the future.

src/leap/mail/imap/mailbox.py
```diff
@@ -514,13 +514,14 @@ def _get_imap_msg(messages):
                 d_imapmsg = []
                 for msg in messages:
                     d_imapmsg.append(getimapmsg(msg))
-                return defer.gatherResults(d_imapmsg)
+                return defer.gatherResults(d_imapmsg, consumeErrors=True)
```
meskio: One day I will understand how `consumeErrors` works. Almost every time I set it up by a try-and-error process, and usually it is the opposite value of what I was expecting.

src/leap/mail/imap/mailbox.py
```diff
@@ -514,13 +514,14 @@ def _get_imap_msg(messages):
...
 
             def _zip_msgid(imap_messages):
                 zipped = zip(
                     list(msg_range), imap_messages)
                 return (item for item in zipped)
 
+            # XXX not called??
```
meskio: Seeing that is unused. Can we just remove the code?

src/leap/mail/imap/messages.py
```diff
@@ -208,6 +208,15 @@ def getSubPart(self, part):
         return IMAPMessagePart(subpart)
 
 
+class CaseInsensitiveDict(dict):
```
meskio: Here (or below where is used) it might be useful to have a comment about the rationale of why we need this dict.

src/leap/mail/imap/mailbox.py
kalikaneko: I see your point, but I also would like to avoid having to lookup what's the meaning of the passed uid == 0 when implementing sequences.

Right now, I'd rather leave it set to false to allow displaying of mail in mutt as a way to cross-check rendering of messages.

Having the first assignment commented is ok for you?

src/leap/mail/imap/mailbox.py
kalikaneko:
```
    @param consumeErrors: (keyword param) a flag, defaulting to False,
    indicating that failures in any of the given L{Deferreds} should not be
    propagated to errbacks added to the individual L{Deferreds} after this
    L{gatherResults} invocation.  Any such errors in the individual
    L{Deferred}s will be converted to a callback result of C{None}.  This                                                                                                    
    is useful to prevent spurious 'Unhandled error in Deferred' messages
    from being logged.  This parameter is available since 11.1.0.
```

src/leap/mail/imap/mailbox.py
kalikaneko: no, the code should be called, but I didn't have the time to focus on where on this round.

src/leap/mail/imap/messages.py
kalikaneko: added both a docstring in the class and a note about its use.

src/leap/mail/imap/mailbox.py
meskio: Yes, I prefer to have it commented, but anyway I'll merge it as it is if you prefer it like that. The ugly hack is there anyway.

src/leap/mail/imap/mailbox.py
meskio: Ok. I'm wondering if we ever have time to look at our `XXX` comments, or we are just piling them up. But fine for me, let's have a look to that in another time.

kalikaneko: that question is isomorphic with "will we ever empty our (chili) backlog?". with the advantage that it's easier to cleanup when touching relating code because stuff doesn't get lost on a huge black hole. I personally have a vim plugin that allows me to navigate the pending tasks when I'm fixing or refactoring a module.

meskio: :) I have similar plugin, I should put the 'TODO' window always visible.

kalikaneko:
force updated with fixes for test regressions, also addressing observations in the review

meskio:
`leap.mail.imap.tests.test_imap.LEAPIMAP4ServerTestCase.testFullAppend` is fixed, but `leap.mail.tests.test_mail.MessageTestCase.test_get_headers` is still a regression at least in my system.

src/leap/mail/adaptors/soledad.py
```diff
@@ -527,26 +530,41 @@ def create(self, store, notify_just_mdoc=False):
...
+            try:
+                d = pending_inserts_dict[msgid]
+                d.callback(msgid)
+            except KeyError:
+                pass
+            return result
+
         # TODO check that the doc_ids in the mdoc are coherent
-        d = []
+        self.d = []
```
meskio: Why do `d` need to become a member of the object? I don't see it used anywhere else.

I don't like having an object member with so sort name, it's even confusing as I will expect a deferred and in reality is a list of deferreds. Can we give it a name that is more meaningful.

PR #179, opened 2015-05-26
post-sync mail processing hooks

kalikaneko:


kalikaneko:
Do not merge yet! Just for review.
this addresses the needs of issue 6996.
This needs https://github.com/leapcode/soledad/pull/221

kalikaneko:
actually, I just realized I need to make some changes:
the set_account method should not be called from this point in the imap service, but from some place in the mail api, so it doesn't get tied to imap service. but then I need to look for the right place to tear down the connected account. I'll do these changes tomorrow, but the overall idea should remain quite similar.

src/leap/mail/sync_hooks.py
```diff
@@ -0,0 +1,112 @@
...
+            self._processing_deferreds = []
+            process_fun = self._queue_doc_id
+
+        for doc_id in doc_id_list:
+            if _get_doc_type_preffix(doc_id) in self.watched_doc_types:
+                log.msg("Mail post-sync hook: processing %s" % doc_id)
+                process_fun(doc_id)
+
+        if self._processing_deferreds:
+            return defer.gatherResults(self._processing_deferreds)
```
meskio: I don't see the need of the if, `gatherResults` can be called with an empty list and will return a Deferred already triggered. And removing this if will remove also the if in `_process_queued_docs`.

meskio:
:+1: to the architecture. Bonus points for keeping the list of unprocessed docs to process them once you have an account configured.

What will happend if I switch accounts whithout restarting accounts and I get an email for the 1st account after I tear down (so there is no account configured) before the 2nd account is set up? Will this email get indexed for the second account? Will that have any implications? We might need to check the order of setUp-tearDown modules in bitmask.

kalikaneko:
> What will happend if I switch accounts whithout restarting accounts

with bitmask, that's currently not possible: you have to stop the services and log-in again, which will replace the soledad object by a new one, and create new imap server instances.

however, things get more tricky when attempting to put the setup/teardown hooks in the generic mail api/service. I have to think more about this.

src/leap/mail/sync_hooks.py
kalikaneko: ah, nice. I'll make that change, thanks.

meskio:
I'm not a big fan of the dummy account idea, I think the thunderbird plugin could do something better than that. But we might need the dummy account as fall back.

For the imap password I liked the idea of having in a file somewhere and the thunderbird plugin picking it up from there.

I have to have a look to cred, but I'm always skeptical of adding yet another abstraction level if we can solve it somehow in a simple manner.

kalikaneko:
done, I've done several fixes.
this also needs https://github.com/leapcode/leap_pycommon/pull/102 apart from the soledad branch.

src/leap/mail/incoming/service.py
```diff
@@ -161,21 +161,15 @@ def fetch(self):
...
-            if failure.check(u1db_errors.InvalidGlobbing):
-                # It looks like we are a dealing with an outdated
-                # mx. Fallback to the version of the index
-                warnings.warn("JUST_MAIL_COMPAT_IDX will be deprecated!",
-                              DeprecationWarning)
-                return self._soledad.get_from_index(
-                    fields.JUST_MAIL_COMPAT_IDX, "*")
-            return failure
+        def _sync_errback(failure):
+            failure.printTraceback()
```
meskio: Could that be done somehow using the log?

src/leap/mail/incoming/service.py
```diff
@@ -246,6 +240,7 @@ def _signal_fetch_to_ui(self, doclist):
         :returns: doclist
         :rtype: iterable
         """
+        # FIXME WTF len(doclist) is 69?
```
meskio: :+1: I was always wondering why is always 69

src/leap/mail/incoming/service.py
kalikaneko: yes: log.err(failure). changed

src/leap/mail/incoming/service.py
kalikaneko: there was a bug here, the callback was passing a list of strings, instead of the expected list of documents. I've added a commit fixing this.

src/leap/mail/sync_hooks.py
meskio: This has not being addressed. I'm ok with the code as it is if you like it as it is, but I like more to remove the two ifs here and in `_process_queued_docs` I find clearer methods that always return the same type of value.

kalikaneko: > if you like it as it is
uh, no, I simply missed it, sorry :(

kalikaneko: added in a new commit

kalikaneko:
added commit for missed refactor @meskio

PR #185, opened 2015-06-30
[bug] return the wrapper, not the soledad doc

kalikaneko:
callbacks on the creation deferred were assuming that the mdoc doc_id
was going to be updated on creation, but here we were returning a
soledad document, not the wrapper itself (that is where the future_id
has been moved to the actual doc_id).

this should fix a problem in which mdocs sometimes do not have a doc_id
to be inserted in the UID hash table. it's related to the sporadic
failure to insert a message when "moving" to a different folder.


meskio:
This pull-req is not solving the issue.

meskio:
Superseeded by #186

PR #187, opened 2015-07-08
[bug] fix the rendering of nested multipart

kalikaneko:
This commit fix a very simplistic and until now broken handling of
nested multipart that went undetected due to the structure of the mails
used in tests until now.  Incidentally, the way that Mail.app structures
attachments made this bug noticeable.

There was also an off-by-one indexing error when retrieving the subpart
message for a given subpart.

Be aware that the current implementation will only handle correctly 2
levels of multipart nesting. Extending beyond in a more generic way will
need further work.

Closes: #7244


src/leap/mail/mail.py
```diff
@@ -190,11 +212,15 @@ def get_subpart(self, part):
 
         sub_pmap = self._pmap.get("part_map", {})
         try:
-            part_map = sub_pmap[str(part + 1)]
+            part_map = sub_pmap[str(part)]
         except KeyError:
-            logger.debug("getSubpart for %s: KeyError" % (part,))
+            log.msg("getSubpart for %s: KeyError" % (part,))
```
meskio: Why no logger?

kalikaneko: to ease debug, I run mail offline with twistd.
it's amazing to see how quickly a pure-reactor app behaves (mainly launching and shutting down)

kalikaneko: Also, for the long run I think we should standarize all the twisty-things to use twisted log facility, it behaves betterfor async logging. And I believe there's been lots of improvements in 15.0

meskio: Then we'll have to see how to integrate them nicely in our logging env as now they look like:
```
[2015-07-08 18:44:15] INFO     - L#109  : twisted.logger._stdlib:__call__ - getSubpart for 9: KeyError
```

While the other logs looks like:
```
[2015-07-08 18:48:10] DEBUG    - L#103  : leap.soledad.client.sync:sync - Soledad sync: there are 0 documents to send.
```

We are loosing information about from where is this log coming, it's severity, ...

kalikaneko: yes, this has been something quick to get the logbook solution capturing them.
I think in the near future we should log specifically from each module (soledad/keymanager/mail) to a file or file-like object, and probably have the centralized logging facility integrating all (something like having tabs, or the ability to see all of them together in the same timeline could be nice)

@ivanalejandro0 some food for thought ^^ in case you want to rescue this for some issue

kalikaneko:
force-pushed. pending review, testing, merge.

src/leap/mail/adaptors/soledad.py
```diff
@@ -688,7 +688,8 @@ def get_body(self, store):
         """
         body_phash = self.hdoc.body
         if not body_phash:
-            return None
+            if self.cdocs:
```
meskio: can you join the two ifs in one `and`?

README.rst
```diff
@@ -24,3 +24,14 @@ fails::
   trial -u leap.mail.imap
 
 Read the *trial* manpage for more options .
+
+imap regressions
+----------------
```
meskio: We have to integrate that in our CI infrastructure as soon as we have one.

src/leap/mail/imap/mailbox.py
```diff
@@ -457,7 +457,33 @@ def expunge(self):
...
+        on the passed uid flag.
+
+        :param uid: If true, the IDs specified in the query are UIDs;
+                    otherwise they are message sequence IDs.
+        :type uid: bool
+        :rtype: callable
+        """
+        get_message_fun = [
+            self.collection.get_message_by_sequence_number,
+            self.collection.get_message_by_uid][uid]
```
meskio: I'm not sure how much I like using a bool as an array index. But if you like your hack I'm happy to keep it here.

kalikaneko: actually, twisted is passing ints (0 or 1). In our code I type it as bool for clarity
Do you see any more elegant solution? an explicit if?

README.rst
kalikaneko: +1000

src/leap/mail/imap/mailbox.py
meskio: I think I would do an explicit if just to keep consistency with the idea of uid being bool, but I'm ok with what you have done.

meskio:
merge-pushed

src/leap/mail/mail.py
ivanalejandro0: I've created a couple of issues to track this stuff:
https://leap.se/code/issues/7247
https://leap.se/code/issues/7262

PR #190, opened 2015-07-23
Revert "Port `enum` to `enum34`"

cz8s:
This reverts commit 16920c065430c9892599b8b803aa6bbd4cd4f933.

this commit never was merged back to develop and it creates merge
conflicts when trying to merge develop into debian/experimental


kalikaneko:
it is not merged into develop because in develop there's no need to use enum anymore.
this is only needed for the 0.3.x branch, and reverting it will make the issue of the incompatibilty resurface.
we're close to deprecate the 0.3.x branch anyhow.

PR #195, opened 2015-08-11
Fix two bugs when converting multipart mails to ContentDocWrappers

fbernitt:
[bug] Fix missing _normailize_dict in DocumentWrapper constructor.

In the constructor values already is normalized (i.e. with underscores),
while kwargs contains items that are not normalized (i.e. with dashes).
Joining the dicts resulted in two entries that only differed by dash or
underscores. The setattr then set the value that occurred later in
items, thereby sometimes overriding the correct value with the default
one.

[bug] Fix typo in content-transfer-encoding in walk.py.

The get_raw_docs method accesses header field content-transfer-encoding
using the string 'content-transfer-type' so the raw doc dict always ends
up with that value set to empty string.

### DIFF
--- a/src/leap/mail/adaptors/models.py
+++ b/src/leap/mail/adaptors/models.py
@@ -76,7 +76,7 @@ class DocumentWrapper(object):
 
         if kwargs:
             values = copy.deepcopy(defaults)
-            values.update(kwargs)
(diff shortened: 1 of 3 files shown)


kalikaneko:
good catch! thanks for the fix and test @fbernitt !

PR #197, opened 2015-09-02
When message is decrypted in IncomingMail, add decrypted header to it

dudadornelles:
We are doing so so that we can know later if that message was originally an encrypted message. We want to tell the user that they can be cool about a message they are reading since that message came in encrypted.

We had that information on X-Leap-Signature for signatures, now we are adding X-Leap-Encryption for encryption stats.

If the header isn't there you can assume the message came in as plain text.

All feedback is more than welcome :)

Thanks,

Duda/Bruno.

### DIFF
--- a/src/leap/mail/incoming/service.py
+++ b/src/leap/mail/incoming/service.py
@@ -465,7 +465,8 @@ class IncomingMail(Service):
         return d
 
     def _add_decrypted_header(self, msg):
-        msg.add_header(self.LEAP_ENCRYPTION_HEADER, self.LEAP_ENCRYPTION_DECRYPTED)
+        msg.add_header(self.LEAP_ENCRYPTION_HEADER,
+                       self.LEAP_ENCRYPTION_DECRYPTED)
 
     def _decrypt_multipart_encrypted_msg(self, msg, encoding, senderAddress):
         """
--- a/src/leap/mail/incoming/service.py
+++ b/src/leap/mail/incoming/service.py
@@ -90,6 +90,7 @@ class IncomingMail(Service):
     CONTENT_KEY = "content"
 
     LEAP_SIGNATURE_HEADER = 'X-Leap-Signature'
+    LEAP_ENCRYPTION_HEADER = 'X-Leap-Encryption'
     """
     Header added to messages when they are decrypted by the fetcher,
     which states the validity of an eventual signature that might be included
@@ -98,6 +99,8 @@ class IncomingMail(Service):
     LEAP_SIGNATURE_VALID = 'valid'
     LEAP_SIGNATURE_INVALID = 'invalid'
     LEAP_SIGNATURE_COULD_NOT_VERIFY = 'could not verify'
+
+    LEAP_ENCRYPTION_DECRYPTED = 'decrypted'
 
     def __init__(self, keymanager, soledad, inbox, userid,
                  check_period=INCOMING_CHECK_PERIOD):
@@ -461,6 +464,9 @@ class IncomingMail(Service):
         d.addCallback(add_leap_header)
         return d
 
+    def _add_decrypted_header(self, msg):
+        msg.add_header(self.LEAP_ENCRYPTION_HEADER, self.LEAP_ENCRYPTION_DECRYPTED)
+
     def _decrypt_multipart_encrypted_msg(self, msg, encoding, senderAddress):
         """
         Decrypt a message with content-type 'multipart/encrypted'.
@@ -503,6 +509,7 @@ class IncomingMail(Service):
 
             # all ok, replace payload by unencrypted payload
             msg.set_payload(decrmsg.get_payload())
+            self._add_decrypted_header(msg)
             return (msg, signkey)
 
         d = self._keymanager.decrypt(
@@ -537,7 +544,9 @@ class IncomingMail(Service):
 
         def decrypted_data(res):
             decrdata, signkey = res
-            return data.replace(pgp_message, decrdata), signkey
+            replaced_data = data.replace(pgp_message, decrdata)
+            self._add_decrypted_header(origmsg)
+            return replaced_data, signkey
 
         def encode_and_return(res):
             data, signkey = res
--- a/src/leap/mail/incoming/tests/test_incoming_mail.py
+++ b/src/leap/mail/incoming/tests/test_incoming_mail.py
@@ -182,7 +182,7 @@ subject: independence of cyberspace
                 self.headers = {}
 
             def add_header(self, k, v):
-                self.headers[k]=v
+                self.headers[k] = v
 
         msg = DummyMsg()
         self.fetcher._add_decrypted_header(msg)
@@ -217,10 +217,7 @@ subject: independence of cyberspace
 
         def add_decrypted_header_called(_):
             self.assertTrue(self.fetcher._add_decrypted_header.called,
-                             "There was some errors with decryption")
-
-
-
+                            "There was some errors with decryption")
 
         d = self._km.encrypt(
             self.EMAIL,
--- a/src/leap/mail/incoming/tests/test_incoming_mail.py
+++ b/src/leap/mail/incoming/tests/test_incoming_mail.py
@@ -176,8 +176,22 @@ subject: independence of cyberspace
         d.addCallback(put_raw_key_called)
         return d
 
+    def testAddDecryptedHeader(self):
+        class DummyMsg():
+            def __init__(self):
+                self.headers = {}
+
+            def add_header(self, k, v):
+                self.headers[k]=v
+
+        msg = DummyMsg()
+        self.fetcher._add_decrypted_header(msg)
+
+        self.assertEquals(msg.headers['X-Leap-Encryption'], 'decrypted')
+
     def testDecryptEmail(self):
         self.fetcher._decryption_error = Mock()
+        self.fetcher._add_decrypted_header = Mock()
 
         def create_encrypted_message(encstr):
             message = Parser().parsestr(self.EMAIL)
@@ -198,8 +212,15 @@ subject: independence of cyberspace
             return newmsg
 
         def decryption_error_not_called(_):
-            self.assertFalse(self.fetcher._decyption_error.called,
+            self.assertFalse(self.fetcher._decryption_error.called,
                              "There was some errors with decryption")
+
+        def add_decrypted_header_called(_):
+            self.assertTrue(self.fetcher._add_decrypted_header.called,
+                             "There was some errors with decryption")
+
+
+
 
         d = self._km.encrypt(
             self.EMAIL,


kalikaneko:
As @elijh pointed out in a recent comment in the discuss list, a X-Leap-Provenance header is already being inserted by the mx when the received message is not encrypted: https://github.com/leapcode/leap_mx/blob/develop/src/leap/mx/mail_receiver.py#L207

So I believe this new header would be redundant, isn't it?
What we need is a way to expose the signature and encryption status for a single message identifier, from the general mail API.

bwagnerr:
Not redundant, because it actually has more information.

I don't think having the mx add this header has much value, it won't have much valuable information (on a basic level I can think of three, not encrypted, could decrypt and could not decrypt, but there can be more depth going forward).

We implemented that as a monkeypatch of leap mail in pixelated, but now with the 0.7 migration we decided to remove it and implement it in leap mail itself. The best place we could find to add such information would be during the decryption process, because that's when you know the result and the details of the decryption itself.

What do you think, @kaliy, @elijh?

kalikaneko:
> it won't have much valuable information (on a basic level I can think of three, not encrypted, could decrypt and could not decrypt

with provenance, the mx states that the message arrived unsigned, and that the mx agent encrypted it with the recipient's pubkey. as @elijh pointed, there's a bug in there yet, and is that we're not removing any previous header.

so, to identify a message that arrived as plaintext, that'd be one with no-leap-provenance.  (assuming the above is fixed).

> If the header isn't there you can assume the message came in as plain text

that's not true. *all* the messages are going to be received encrypted in the client-side at the moment in which they are processed from the incoming queue. only the lack of provenance header will tell you if the message arrived to the mx in plaintext or not.

that said, it's true that we need to make patent to the user if what he's seeing was plaintext originally or not. I think the info has to come from the mail api, not be elucidated at the mua, and there must be some way of clearly asking for encryptiong/signature combinations to the api.

Coming back to the semantics of your proposed header:

> not encrypted

if you're meaning "not encrypted when received at mx", this you cannot know without checking the leap-provenance header, see above.

> could decrypt

I do see a value about adding some "decrypted" metadata. mainly for timestamping the act of decryption, although one should be careful about whether this is an information that we want to store as a mail header or not. I think I would store it as metadata in some soledad document, not into a header (think about information leakage when forwarding, for instance).

I also think it's interesting to differentiate between mail that we originally processed (ie, coming via mx and soledad sync) and mail that we could have inserted into a folder from some other procedence (for instance, pasting a message from another account in the MUA).

> and could not decrypt

why would this happen?
having being encrypted with a wrong key, I guess.
I think this would be self-evident, having the right content-type and a gpg block in the payload.

> but there can be more depth going forward

let's think about those.

kalikaneko:
@meskio i'd like to hear your opinions on this one too. I think we need to schedule the signature/encryption api for mail 0.4.1, which should closely follow 0.4.0 release.

elijh:
yes, by the time leap_mail sees any incoming message, it should be encrypted to the user's private key. If it is ever *not* the case, then this is a grave error, but probably not something that needs to be reported to the MUA via a header. 

The only way that leap_mail has of knowing if the message was end-to-end encrypted is to check the signature of encrypted messages (which must be distinguished from signatures made by the sender on cleartext messages).

So, there *is* a problem, and a bug, that needs to be addressed: leap_mx must add different headers for messages that are encrypted+signed, versus messages that are signed, and then encrypted by leap_mx. Currently, messages of the latter sort are not processed correctly, but... if they were, we need to make sure that the headers are different than in the case of encrypted+signed.

So, there are four possible states an email can be in...

(e = encrypted, s = signed, 1 = yes, 0 = no)

this is what currently happens (in my tests):

e1s1
* X-Leap-Signature: valid; pubkey="8688B48800440025"

e0s1
* X-Leap-Provenance: Mon, 24 Aug 2015 21:02:04 -0000; pubkey="32D445F0E76ED2C9"
* X-Leap-Signature: could not verify

e1s0
* X-Leap-Signature: invalid

e0s0
* X-Leap-Provenance: Mon, 24 Aug 2015 21:02:57 -0000; pubkey="32D445F0E76ED2C9"
* X-Leap-Signature: could not verify

Technically, for e0s1, the signature should validate. It would be good for leap_mail to include an indication that it *only* validated the signature--the signature was not on the encrypted message but was a cleartext signature. So, instead of "valid", it could say "valid (cleartext)" or something. The existence of the x-leap-provenance header tells the mua the same information, but relies on leap_mx for this when leap_mail is also able to figure it out, and is a more trusted source of information.

I think that one of the main problems is that all this is complicated and confusing. It would be much easier to have a "X-Leap-Encrypted" header that clearly stated what is what. The problem with this is that e1s1 is the only state for which X-Leap-Encrypted should ever be true, since e1s0 might actually *not* have been encrypted by the sender, but if e1s1 is the only time we would use X-Leap-Encrypted, then it is redundant because X-Leap-Signature already gives us this information.

However, in the interest of making the MUA easier to write (thunderbird extension and pixelated ui), it may make a lot of sense to shield the MUA from any of this logic and have a header that just tells the MUA what should be shown. So, how about a X-Leap-Secure header, which is just "yes" or "no":

e1s1 => yes
e0s1 => no
e1s0 => yes, but From header is replaced with "Anonymous (alice@example.org)" this is to let people know that they cannot trust the identity of the sender, even though the message is encrypted.
e0s0 => no

This only makes sense if leap_mx will only consider the signature valid if it is from a key that has a uid that matches the from email address. Otherwise, it would be trivially easy to send a "secure" message impersonating anyone else.

That signature to from header validation is not happening currently, which to me seems like a high priority (!)

bwagnerr:
@elijh it became clearer to me when talking to @kalikaneko today that we might be talking about different things, it looks like we have encryption for incoming mail blobs (incoming mail is encrypted to be stored and later processed at the client leap mail queue) and we have PGP mail encryption. The stage where we added verification for encryption status is after the mail arrived at the client and has been succesfully blob decrypted, in that case we have the original mail that the provider received and we can determine if it was encrypted, signed, both or none and get details about those. The mx encrypts the raw mail blob, so I don't see a chance of those conflicting of reporting erroneously.

Also another point is that I would really like to have signature and encryption separately, instead of a "secure" or "not secure" because that way the MUA can choose how to best handle the information and present to the user.

What do you think?

elijh:
@bwagnerr I think we are not talking about different things. Lets define some words to make the discussion easier:

* encryption wrapper: the encryption that leap_mx does of *all* incoming email, encrypting the entire message, headers and all.
* message encryption: the OpenPGP encryption of the message that was created by the original sender.
* MUA: the UI for email, either pixelated webmail ui or thunderbird, etc.

As I understand it, you would like leap.mail to be able to clearly and definitively say to MUA if a particular message had 'message encryption'.

To complicate things, I think that maybe it used to be the case the leap_mx did add not the encryption wrapper but would add message encryption if message was not encrypted. We don't want it to do this, except that sometimes we do (when the message is to be forwarded on to another address). The other address might very well be a LEAP-powered address, and we don't want to trick the final recipient into thinking the email was end-to-end encrypted. (true, we don't have forwards working yet, but planning ahead...)

The second complication is that leap.mail can *never* know definitively if there was message encryption or not, unless the message was properly signed by the sender's key.

So! For the forwarding reasons, I think it is essential to keep the X-Leap-Provenance headers, even though it seems unless at first.

The second question is, should leap.mail add a X-Leap-Encrypted header? I still maintain that such a header will add no new information. The reason for this claim is that leap.mail cannot trust unsigned message encryption, because any MiTM could have added it, including sender provider or recipient provider. 

But, I do see why it is confusing to not have X-Leap-Encrypted. How about this as a compromise?

e1s1:

* X-Leap-Signature: valid
* X-Leap-Encryption: valid

e1s0:

* X-Leap-Signature: none
* X-Leap-Encryption: maybe

e0s1:

* X-Leap-Signature: valid
* X-Leap-Encryption: none

e0s0:

* X-Leap-Signature: none
* X-Leap-Encryption: none

Does that make sense? What I *don't* want to do is say that e1s0 has valid encryption, because this is actually totally unknown.

meskio:
> What I don't want to do is say that e1s0 has valid encryption, because this is actually totally unknown.

In this case e1s1 is also unknown, if a e0s1 is forwarded by encrypting it will become a e1s1 but the original message was not encrypted. So if you don't consider a valid encryption if is not signed I don't see why you consider it if is singed.

There is more weird corner cases for forwarding. If the email is forwarded to a leap provider this provider will remove the X-Leap-Provenance header (or this is what we are proposing in the mailing list), and add it's own.

I need to think more about the forwarding use case.

cz8s:
maybe we should fix the existing situation and think about forwarding if we have something working :)

elijh:
> > What I don't want to do is say that e1s0 has valid encryption, because this is actually totally unknown.
> In this case e1s1 is also unknown, if a e0s1 is forwarded by encrypting it will become a e1s1 but the original message was not encrypted. So if you don't consider a valid encryption if is not signed I don't see why you consider it if is singed.

Well, that is a very good point. 

> maybe we should fix the existing situation and think about forwarding if we have something working

The core issue is not about forwarding. Let me start over: I will describe the attack that has me really worried and you all can say if you think this attack is something we should work to prevent, and if so, how to prevent it.

Suppose you are a human rights lawyer in a country where the government is likely trying to read your emails. You sign up with a LEAP-powered provider, and are happily using encrypted email. In particular, it is very important to you to be able to know that certain emails have been end-to-end encrypted.

Suddenly, you start to receive a lot of encrypted emails from your contacts that warn you that one of your coworkers is a government spy. You become very paranoid and no longer share any information with this coworker.

In this scenario, it was the government that sent all the encrypted emails. This is very easy to do, anyone can discover your public key and send you an encrypted email. As long as they don't sign the message, and the domain does not have SPF or DKIM, the message will validate as a perfectly valid encrypted email from whomever the attacker wants it to be from.

Basically, I am very worried about giving users a false sense of security when an email is encrypted but not signed. I think that encrypted but not signed doesn't mean shit. If we ever give users the impression that these kinds of emails are anything more than shit, then we are inviting attacks of this nature.

Yes, it is also true that an attacker can take any cleartext signed message and encrypt it, but this attack does not trouble me as much. In this case, the attacker tricked the recipient into thinking that a message was confidential when it was not, which is serious, but seems less likely to happen (since it requires MiTM and software such as enigmail defaults to turning on encryption when replying to an encrypted message).

There is actually a way to prevent this latter attack, since OpenPGP allows you to encrypt then sign instead of sign then encrypt, but no one does it this way because it exposes the sender (you can also sign, then encrypt, then sign again, if you want). If using memory hole, I think there might be an easier way (include hash of symmetric encryption session key in the encrypted headers), but I will ask dkg about that idea.

cz8s:
We should not put usability logic into leap-mx. If you are afraid that someone could be mislead by mixing up signing and encrypting, you need to tackle that problem at the ui level. leap-mail can not tell the user what to do. leap-mail should tell the MUA what status the messages had and the MUA then should tell the user in a way that the user can make informed descisions.

meskio:
I agree with @cz8s, @elijh is talking about a usability issue, we should worry about usability issues in the MUA. I think it makes sense to signal correctly the MUA about the status of each message, and the MUA should display the status how it better fits, and here I agree that the MUA should not signal an email as secure if is not signed. But I don't understand how this attack relates to have a proper way to signal what is happening.

Talking with @kalikaneko we've agree that the headers have other issues, as you could inject emails via IMAP and get wrong headers. We could do our best to clean up headers to avoid this attacks. But it might be worth it for the long term to have other way (and API in leap.mail and maybe a websockets channel for thunderbird) to signal the status of the email. The first step on that could be the issue https://leap.se/code/issues/6914

Anyway I think we should merge this PR and start working on an API to make headers obsolete.

cz8s:
+1. The encryption status should be moved from the headers to something more practical.
But for now they work and we should start with that and then make it better :)

elijh:
> We should not put usability logic into leap-mx

I think you mean leap.mail?

The fundamental question here is what type of assumptions will users make when we say that a message is secure or present a particular sender. The OpenPGP spec has a discussion of this, iirc, and it basically boils down to this: In order to use OpenPGP correctly, the user needs to understand how email works at a very technical level. 

We are obviously trying for something very different. There are no issues that are "merely" for the UI to figure out, because certain decisions about how information is presented cannot be taken lightly. I am convinced that the logic required to properly present to the user reasonable assertions is going to be (a) more complex than should be in the UI layer, and (b) require access to crypto functions not available in the UI layer. 

I will work on a more complete proposal of what this logic could look like, and how this should be communicated to the UI layer. In the mean time, I don't want to hold up work in other areas. So, please merge this now if this change is needed by other code, with the understanding that this approach can only be temporary because a simple signal of encrypted and/or signed is totally inadequate for the long run.

meskio:
Merged, let's keep talking in other channels about a mid-term solution with the proposals from @elijh.

PR #199, opened 2015-09-15
[feat] use async events api

kalikaneko:
in this way, we're using twisted reactor instead of having another
thread with zmq's own copy of tornado ioloop.

Resolves: #7274


meskio:
I rebase merged. I lost the signature by doing that :(

kaeff:
@kalikaneko @meskio We're seeing test failures on our fork's CI: https://snap-ci.com/pixelated/soledad/branch/develop/logs/defaultPipeline/108/Tests

Any idea?

PR #202, opened 2015-09-22
[bug] don't extract openpgp header if valid attached key

meskio:
The key extracto should check first for attached keys and if this fails
then should try the OpenPGP header.

- Resolves: #7480


kalikaneko:
fix typo in commit message (extracto)

PR #204, opened 2015-09-22
[refactor] log the added key explicitely

kalikaneko:


src/leap/mail/incoming/service.py
```diff
@@ -695,6 +695,9 @@ def _maybe_extract_attached_key(self, attachments, address):
         """
         MIME_KEY = "application/pgp-keys"
 
+        def log_key_added(ignored):
+            logger.debug('Added key found in attachment for %s' % address)
```
meskio: It should `return True` or the OpenPGP header will be parsed as well.

kalikaneko: auch! fixin...

PR #214, opened 2015-12-04
[feat] credentials handling: use twisted.cred

kalikaneko:
==Work in progress.==
This will be needed to try the latest bonafide/bitmask.core stuff.
This will also need some adjustments in the current initialization in the bitmask client.

### DIFF
--- a/src/leap/mail/imap/account.py
+++ b/src/leap/mail/imap/account.py
@@ -49,6 +49,7 @@ if PROFILE_CMD:
 # Soledad IMAP Account
 #######################################
 
+
 class IMAPAccount(object):
     """
     An implementation of an imap4 Account
@@ -68,8 +69,8 @@ class IMAPAccount(object):
         You can either pass a deferred to this constructor, or use
         `callWhenReady` method.
 
-        :param user_id: The name of the account (user id, in the form
-                        user@provider).
+        :param user_id: The identifier of the user this account belongs to
+                        (user id, in the form user@provider).
         :type user_id: str
 
         :param store: a Soledad instance.
--- a/src/leap/mail/imap/server.py
+++ b/src/leap/mail/imap/server.py
@@ -20,15 +20,9 @@ LEAP IMAP4 Server Implementation.
 import StringIO
 from copy import copy
 
-from twisted import cred
-from twisted.internet import reactor
 from twisted.internet.defer import maybeDeferred
 from twisted.mail import imap4
 from twisted.python import log
-
-from leap.common.check import leap_assert, leap_assert_type
-from leap.common.events import emit_async, catalog
-from leap.soledad.client import Soledad
 
 # imports for LITERAL+ patch
 from twisted.internet import defer, interfaces
@@ -72,25 +66,6 @@ class LEAPIMAPServer(imap4.IMAP4Server):
     """
     An IMAP4 Server with a LEAP Storage Backend.
     """
-    def __init__(self, *args, **kwargs):
-        # pop extraneous arguments
-        soledad = kwargs.pop('soledad', None)
-        uuid = kwargs.pop('uuid', None)
-        userid = kwargs.pop('userid', None)
-
-        leap_assert(soledad, "need a soledad instance")
-        leap_assert_type(soledad, Soledad)
-        leap_assert(uuid, "need a user in the initialization")
-
-        self._userid = userid
-
-        # initialize imap server!
-        imap4.IMAP4Server.__init__(self, *args, **kwargs)
-
-        # we should initialize the account here,
-        # but we move it to the factory so we can
-        # populate the test account properly (and only once
-        # per session)
 
     #############################################################
     #
@@ -181,10 +156,6 @@ class LEAPIMAPServer(imap4.IMAP4Server):
         :param line: the line from the server, without the line delimiter.
         :type line: str
         """
-        if self.theAccount.session_ended is True and self.state != "unauth":
-            log.msg("Closing the session. State: unauth")
-            self.state = "unauth"
-
         if "login" in line.lower():
             # avoid to log the pass, even though we are using a dummy auth
             # by now.
@@ -208,24 +179,6 @@ class LEAPIMAPServer(imap4.IMAP4Server):
             self.mbox = None
         self.state = 'unauth'
 
-    def authenticateLogin(self, username, password):
-        """
-        Lookup the account with the given parameters, and deny
-        the improper combinations.
-
-        :param username: the username that is attempting authentication.
-        :type username: str
-        :param password: the password to authenticate with.
-        :type password: str
-        """
-        # XXX this should use portal:
-        # return portal.login(cred.credentials.UsernamePassword(user, pass)
-        if username != self._userid:
-            # bad username, reject.
-            raise cred.error.UnauthorizedLogin()
-        # any dummy password is allowed so far. use realm instead!
-        emit_async(catalog.IMAP_CLIENT_LOGIN, username, "1")
-        return imap4.IAccount, self.theAccount, lambda: None
 
     def do_FETCH(self, tag, messages, query, uid=0):
         """
--- a/src/leap/mail/imap/service/imap-server.tac
+++ b/src/leap/mail/imap/service/imap-server.tac
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # imap-server.tac
 # Copyright (C) 2013,2014 LEAP
@@ -27,6 +28,9 @@ userid = 'user@provider'
 uuid = 'deadbeefdeadabad'
 passwd = 'supersecret' # optional, will get prompted if not found.
 """
+
+# TODO -- this .tac file should be deprecated in favor of bitmask.core.bitmaskd
+
 import ConfigParser
 import getpass
 import os
@@ -112,7 +116,7 @@ tempdir = "/tmp/"
 
 print "[~] user:", userid
 soledad = initialize_soledad(uuid, userid, passwd, secrets,
-                             localdb, gnupg_home, tempdir)
+                             localdb, gnupg_home, tempdir, userid=userid)
 km_args = (userid, "https://localhost", soledad)
 km_kwargs = {
     "token": "",
@@ -131,7 +135,8 @@ keymanager = KeyManager(*km_args, **km_kwargs)
 
 
 def getIMAPService():
-    factory = imap.LeapIMAPFactory(uuid, userid, soledad)
+    soledad_sessions = {userid: soledad}
+    factory = imap.LeapIMAPFactory(soledad_sessions)
     return internet.TCPServer(port, factory, interface="localhost")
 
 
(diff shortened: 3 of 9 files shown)


kalikaneko:
@ivanalejandro0 you will need this if you want to test latest core/cli

src/leap/mail/imap/account.py
```diff
@@ -49,6 +49,16 @@ def _debugProfiling(result, cmdname, start):
...
 #######################################
 
+class IIMAPAccount(Interface):
+    """
+    The abstraction describing an IMAPAccount.
+    """
+
+    user_id = Attribute("""The identifier for the account (user@provider)""")
+    account = Attribute("An internal reference to a leap.mail Account, "
+                        "which is initialized with a Soledad instance""")
```
meskio: Too many closing quotes.

src/leap/mail/imap/service/imap.py
```diff
@@ -41,57 +47,129 @@
...
+                    avatar = IMAPAccount(avatarId, soledad)
+                    return (IAccount, avatar,
+                            getattr(avatar, 'logout', lambda: None))
+            raise NotImplementedError(self, interfaces)
+
+        return self.lookupSoledadInstance(avatarId).addCallback(gotSoledad)
+
+    def lookupSoledadInstance(self, userid):
+        soledad = self._soledad_sessions[userid]
+        return defer.succeed(soledad)
```
meskio: Why add a Deferred here? I think the code of `requestAvatar` could be simpler if this is not a Deferred.

src/leap/mail/imap/service/imap.py
```diff
@@ -15,22 +15,28 @@
...
 from twisted.internet.protocol import ServerFactory
-from twisted.mail import imap4
 from twisted.python import log
+from twisted.python.failure import Failure
+from zope.interface import implementer
 
 from leap.common.events import emit_async, catalog
-from leap.common.check import leap_check
-from leap.mail.imap.account import IMAPAccount
+from leap.mail.imap.account import IMAPAccount, IIMAPAccount
```
meskio: I don't see IIMAPAccount used anywhere.

src/leap/mail/imap/service/imap.py
```diff
@@ -15,22 +15,28 @@
...
+from twisted.cred.checkers import ICredentialsChecker
+from twisted.cred.error import UnauthorizedLogin
+from twisted.mail.imap4 import IAccount
+from twisted.internet import defer
 from twisted.internet import reactor
 from twisted.internet.error import CannotListenError
 from twisted.internet.protocol import ServerFactory
-from twisted.mail import imap4
 from twisted.python import log
+from twisted.python.failure import Failure
```
meskio: I don't see Failure used anywhere.

src/leap/mail/imap/service/imap.py
```diff
@@ -164,6 +218,7 @@ def run_service(store, **kwargs):
         if os.environ.get("LEAP_DOCKERIZED"):
             interface = ''
 
+        # TODO use Endpoints !!!
```
meskio: What are `Endpoints`?

src/leap/mail/imap/account.py
```diff
@@ -49,6 +49,16 @@ def _debugProfiling(result, cmdname, start):
 # Soledad IMAP Account
 #######################################
 
+class IIMAPAccount(Interface):
+    """
+    The abstraction describing an IMAPAccount.
+    """
```
meskio: I guess this is somehow related to `imap4.IAccount`, isn't it? I'm not sure if `zope.interface` has something to state this relation, but maybe we can mention it in the docstring.

src/leap/mail/imap/account.py
```diff
@@ -49,6 +49,16 @@ def _debugProfiling(result, cmdname, start):
...
+
+    user_id = Attribute("""The identifier for the account (user@provider)""")
+    account = Attribute("An internal reference to a leap.mail Account, "
+                        "which is initialized with a Soledad instance""")
+
+
 class IMAPAccount(object):
```
meskio: Should this class `@implementer(IAccount)` (or IIMAPAccount). I see it in `LocalSoledadIMAPRealm` is used as IAccount, I guess it will be nice to make the relation more explicit.

src/leap/mail/imap/server.py
```diff
@@ -73,25 +73,8 @@ class LEAPIMAPServer(imap4.IMAP4Server):
     An IMAP4 Server with a LEAP Storage Backend.
```
meskio: There is a bunch of unused dependencies in this file: cred, reactor & leap.*

src/leap/mail/imap/server.py
```diff
@@ -73,25 +73,8 @@ class LEAPIMAPServer(imap4.IMAP4Server):
...
-        userid = kwargs.pop('userid', None)
-
-        leap_assert(soledad, "need a soledad instance")
-        leap_assert_type(soledad, Soledad)
-        leap_assert(uuid, "need a user in the initialization")
-
-        self._userid = userid
-
-        # initialize imap server!
         imap4.IMAP4Server.__init__(self, *args, **kwargs)
```
meskio: Maybe we don't need to have an `__init__` in this class.

src/leap/mail/imap/service/imap.py
```diff
@@ -103,13 +181,8 @@ def buildProtocol(self, addr):
         # TODO should reject anything from addr != localhost,
         # just in case.
         log.msg("Building protocol for connection %s" % addr)
-        imapProtocol = self.protocol(
-            uuid=self._uuid,
-            userid=self._userid,
-            soledad=self._soledad)
-        imapProtocol.theAccount = self.theAccount
+        imapProtocol = self.protocol(self._soledad_sessions)
         imapProtocol.factory = self
```
meskio: Do `LocalSoledadIMAPServer` need to know about the factory? I think is not used anywhere.

src/leap/mail/imap/service/imap-server.tac
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # imap-server.tac
 # Copyright (C) 2013,2014 LEAP
```
meskio: The `imap.LeapIMAPFactory` initialization it's still passing the old params.

src/leap/mail/imap/service/imap.py
```diff
@@ -41,57 +47,129 @@
...
     capabilities.
     """
-    protocol = LEAPIMAPServer
 
-    def __init__(self, uuid, userid, soledad):
+    protocol = LocalSoledadIMAPServer
+
+    def __init__(self, soledad_sessions):
         """
         Initializes the server factory.
```
meskio: As this will be the entry point from outside maybe we should document what `soledad_sessions` are here.

src/leap/mail/imap/service/imap.py
```diff
@@ -146,16 +208,8 @@ def run_service(store, **kwargs):
     :returns: the port as returned by the reactor when starts listening, and
```
meskio: docstring mentions store, maybe it's good to document `soledad_sessions` hear (and maybe is not needed to do it in `LeapIMAPFactory`).

src/leap/mail/imap/service/imap.py
```diff
@@ -146,16 +208,8 @@ def run_service(store, **kwargs):
...
               the factory for the protocol.
     """
-    leap_check(store, "store cannot be None")
-    # XXX this can also be a ProxiedObject, FIXME
-    # leap_assert_type(store, Soledad)
-
     port = kwargs.get('port', IMAP_PORT)
```
meskio: I would remove the `**kwargs` from the params and add `port = IMAP_PORT` to make it explicit.

meskio:
It looks good. It reduces the interdependency of classes. Most of the `__init__` methods now have less parameters and there is less injection of class variables.

src/leap/mail/imap/service/imap.py
kalikaneko: https://twistedmatrix.com/documents/current/core/howto/endpoints.html
nowadays it's preferred to use endpoints (interfaces with a single method: listen) basically, rather than calling listenTCP direclty.

src/leap/mail/imap/account.py
kalikaneko: I can get rid of this; I thought at some point that I needed a specific interface for the avatar.

src/leap/mail/imap/account.py
kalikaneko: removing IIMAPAccount, it was not used after some refactor.

src/leap/mail/imap/service/imap.py
kalikaneko: that is true.
however, IRealm.requestAvatar expects a deferred.

also, I just realized that, ideally, we should return the soledad instance just after all the initialization callbacks have been initialized (we don't do right now, but we could move the whenReady callbacks to inside soledad client api).

src/leap/mail/imap/service/imap.py
kalikaneko: I stopped using it (request avatar checks IAccount becase that's what the imap server deals with), so it wasn't really needed. removed the rests of it.

src/leap/mail/imap/service/imap.py
kalikaneko: true, deleted.

src/leap/mail/imap/account.py
kalikaneko: deleted, see above.

src/leap/mail/imap/server.py
kalikaneko: fixed!

src/leap/mail/imap/server.py
kalikaneko: uh, that's right. fixed.

src/leap/mail/imap/service/imap.py
kalikaneko: good catch! I think it was needed before for some hack. not anymore, removed.

src/leap/mail/imap/service/imap-server.tac
kalikaneko: fixed, thanks. I think with the new bitmask.core.bitmaskd we can deprecate this .tac, its use was mainly to be able to instantiate a local imap server quickly.

src/leap/mail/imap/service/imap.py
kalikaneko: true. added the param docstring to several places.

src/leap/mail/imap/service/imap.py
kalikaneko: documented in both places. my idea is to bring here the mail multi-services (they are now in bitmask_core.mail_services), so those would be the main entrypoints. run_service would be a "private" method then.

src/leap/mail/imap/service/imap.py
kalikaneko: good idea, done.

kalikaneko:
rebased, squashed, signed and pushed.

PR #215, opened 2015-12-10
Fixed the get_body logic

bwagnerr:
The rest of the code is not prepared to deal with a None body, this used to happen when a mail doesn't have a body phash neither cdocs and it would break leap mail (this happens in unexpected cases, a mail should always have a body, although could be a blank one, but sometimes the body document was not synced yet or couldn't be found)

In those cases, we return an empty body, so leap mail can still work normally and process the rest of the mails and requests without breaking.


kalikaneko:
merged.
it's strange because I don't see a694fe4 on the merged branch. I assume it's some GH glitch.

PR #217, opened 2015-12-16
[feat] cred authentication for SMTP service

kalikaneko:


src/leap/mail/smtp/__init__.py
```diff
@@ -26,44 +26,33 @@
 from leap.mail.outgoing.service import OutgoingMail
```
meskio: Not used

src/leap/mail/smtp/gateway.py
```diff
@@ -245,6 +311,7 @@ def validateFrom(self, helo, origin):
         @raise twisted.mail.smtp.SMTPBadSender: Raised if messages from this
             address are not to be accepted.
         """
+        print "VALIDATE FROM 2", helo, origin
```
meskio: I guess this is print can be removed.

src/leap/mail/smtp/__init__.py
kalikaneko: done

src/leap/mail/smtp/gateway.py
kalikaneko: print 'done'

meskio: ???

kalikaneko: hmm I thought I had removed the print, but it seems I didn't. pushed again.

PR #218, opened 2015-12-21
[feat] use fingerprint instead of key_id to address keys

meskio:
Related to https://github.com/leapcode/keymanager/pull/94


kalikaneko:
rebased & merged

PR #219, opened 2016-01-05
Fix pixelated repos, u1db and dirspec reference on requirements

giovaneliberato:
The packages for u1db and dirspec were removed from pypi. Now we need to download it directly from canonical's launchpad.
Also, updates pixelated repositories references.


kalikaneko:
merged #227 instead

PR #220, opened 2016-01-21
Process and verify signature on plain text mails

giovaneliberato:
On pixelated we had the need to show information on plain text signed mail, but the leap mail didn't support that yet.

We had to create a new Generator due to the fact that python's `email.generator.Generator` would remove some `\n` from headers and the signature check would fail.

Please reach out if you have any comments or suggestions.

Pairing with @jeffhsta


src/leap/mail/generator.py
```diff
@@ -0,0 +1,20 @@
...
+    
+    This code was extracted from Mailman.Generator.Generator, version 2.1.4:
+    
+    Most other Generator will be created not setting the foldheader flag,
+    as we do not overwrite clone(). The original clone() does not
+    set foldheaders.
+
+    So you need to set foldheaders if you want the toplevel to fold headers
+    This class was extracted from
+    """
```
meskio: Can you add a XXX or TODO comment mentioning the version of python that solves this issue.

src/leap/mail/incoming/service.py
```diff
@@ -522,8 +523,8 @@ def build_msg(res):
             return (msg, signkey)
 
         d = self._keymanager.decrypt(
-            encdata, self._userid, OpenPGPKey,
-            verify=senderAddress)
+                encdata, self._userid, OpenPGPKey,
+                verify=senderAddress)
```
meskio: Why changing the indentation here? I think the new indentation is not pep8.

src/leap/mail/incoming/service.py
```diff
@@ -393,9 +393,7 @@ def _process_decrypted_doc(self, doc, data):
             defer.returnValue(False)
 
         # ok, this is an incoming message
-        rawmsg = msg.get(self.CONTENT_KEY, None)
-        if not rawmsg:
-            return ""
+        rawmsg = msg.get(self.CONTENT_KEY, "")
```
giovaneliberato: Should I revert this change? Just realized that it was returning the empty string straight forward instead of passing to the _maybe_decrypt_msg function.

src/leap/mail/incoming/service.py
giovaneliberato: Editor problems.. I'll fix this.

src/leap/mail/generator.py
giovaneliberato: Sure!

src/leap/mail/incoming/service.py
meskio: I guess the change is innocuous. But let's revert it just in case.

giovaneliberato:
This is already fixed on python 3.3

I'm pasting the link here for future reference https://bugs.python.org/issue968430

src/leap/mail/incoming/service.py
```diff
@@ -578,6 +578,45 @@ def encode_and_return(res):
...
+
+        :return: A Deferred that will be fired with a tuple containing a
+        signed Message and the signing OpenPGPKey if the signature
+        is valid or InvalidSignature.
+        :rtype: Deferred
+        """
+        msg = copy.deepcopy(origmsg)
+        data = self._serialize_msg(msg.get_payload(0))
+        detached_sig = self._extract_signature(msg)
+        d = self._keymanager.verify(data, sender_address, OpenPGPKey, detached_sig)
```
meskio: Another pep8 error: "line too long"

meskio:
If you don't mind squashing the last commit in the previous ones to make the history cleaner, I'll be happy to merge it.

giovaneliberato:
@meskio squash done. 
Just a question: How many chars does leap use to identify lines too long? 80?

meskio:
I'm using the pep8 tool to check it. I think is 80 chars.

meskio:
I've fixed the pep8 and merge it.

PR #221, opened 2016-02-02
[bug] Use the right succeed function for passthrough encrypted email

meskio:
- Resolves #7861

### DIFF
--- a/changes/next-changelog.rst
+++ b/changes/next-changelog.rst
@@ -19,9 +19,10 @@ Features
 
 Bugfixes
 ~~~~~~~~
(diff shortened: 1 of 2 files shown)


KwadroNaut:
Works fine, good fix, please merge. Intentional that when encrypted but not signed, that it is passed through as is?

meskio:
Yes, it passes the message through if is encrypted. If it's not encrypted (doesn't matter if it's signed) it will encrypt automatically and I think bitmask will add a second signature with it's own key (I will need to check the code to be sure).

kalikaneko:
rebased & merged

PR #224, opened 2016-02-10
[feat] Use cryptography instead of pycryptopp to reduce dependencies.

meskio:
* Resolves: #7889

### DIFF
--- a/src/leap/mail/walk.py
+++ b/src/leap/mail/walk.py
@@ -17,13 +17,16 @@
 """
 Utilities for walking along a message tree.
 """
-from pycryptopp.hash import sha256
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes
 
 from leap.mail.utils import first
 
 
 def get_hash(s):
-    return sha256.SHA256(s).hexdigest()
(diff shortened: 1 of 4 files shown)


src/leap/mail/walk.py
```diff
@@ -19,18 +19,24 @@
...
+
+def _get_hash(s):
+    digest = hashes.Hash(hashes.SHA256(), default_backend())
+    digest.update(s)
+    return digest.finalize().encode("hex").upper()
+
 if DEBUG:
     def get_hash(s):
-        return sha256.SHA256(s).hexdigest()[:10]
+        return _get_hash(s)[:10]
```
meskio: Why there is this debug function?

kalikaneko: it's a remnant from when this module was in development, it shows and print different parts of the message.

meskio: Can we remove it?

kalikaneko: I'm ok with removing it, do you mind to include it in the pr? 
note there're several refs to DEBUG in the module

meskio: Sure, let me update it.

PR #227, opened 2016-03-14
[Fix] Fix pixelated repos reference on requirements

cz8s:
Fix requirements


kalikaneko:
can you remove the pycryptopp commit?

PR #228, opened 2016-03-14
Improved imapmailbox listeners

cz8s:
make IMAPMailbox listeners deferred and support multiuser


bwagnerr:
That's because we use IMAPAccount to access mailboxes, listen to new ones and such. We've been using it since we switched to using leap mail instead of the docs directly.

kalikaneko:
>  We've been using it since we switched to using leap mail instead of the docs directly.

ok, but then something says to me that the abstraction barrier between Account / IMAPAccount is broken. what exactly is there in IMAPAccount that cannot be used from Account directly? addListener etc are methods of Account.

src/leap/mail/imap/account.py
```diff
@@ -244,15 +249,15 @@ def select(self, name, readwrite=1):
         """
         name = normalize_mailbox(name)
 
+        if name in self._mailboxes:
+            return defer.succeed(self._mailboxes[name])
```
kalikaneko: I see where the problem is coming from, but I don't think this is a proper solution.
If several MUAs are manipulating the same mailbox, in the IMAP sense, what one MUA does can interfere with the others. For instance, a select with writeable false, will set the state for another MUA using it.

I'm trying to find a better solution for the global dict of listeners.

src/leap/mail/imap/mailbox.py
```diff
@@ -420,15 +420,22 @@ def notify_new(self, *args):
...
         def cbNotifyNew(result):
             exists, recent = result
+
+            results = []
             for listener in self.listeners:
-                listener.newMessages(exists, recent)
+                d = defer.maybeDeferred(listener.newMessages, exists, recent)
+                results.append(d)
+            yield defer.gatherResults(results)
+            defer.returnValue(result)
```
kalikaneko: do you need to wait for the notification?

src/leap/mail/imap/account.py
kalikaneko: when reviewing this, I realized that the status of notifications was actually broken for us.
it was complicated by the fact that we had two different places in which we instantiated a top-level Account object, so the listener set for the underlying message_collection wasn't shared.

have a look at commit https://github.com/leapcode/leap_mail/pull/231/commits/ab50358222db1d78ba8384ac299443b4c95397ee and see if this would work for your case.

kalikaneko:
I have two main concerts with this:

1. re-using IMAPmailbox instances is not feasible. little, but there's some state that shouldn't be shared since they represent views.
2. the global dict is wrong in multi-user scenario. that was mainly a hack for the fact that, in some situations, this dict only grows but is never trimmed (stress-testing, for instance), and so the time taken to notify all listeners grows with time. a size-limited dict/set should be used instead (and probably storing the time also, so that older connections can be removed).
3. underlying message-collections are cached in the account object (and with the proposed commit, indexed by userid), so maybe it's not needed to cache the mailboxes too.

bwagnerr:
I tried to swap ImapAccount for Account and the ready callback stopped being called. That broke all the tests and the user agent itself.

But I agree we should move to Account at some point, though we have to investigate that further how to do this and we don't have enough hands now

bwagnerr:
Kaliy, these changes are needed for us to run the user-agent multiuser for now. Do you think these won't be needed with your latest changes? If not what would you propose we could do?

kalikaneko:
> I tried to swap ImapAccount for Account and the ready callback stopped being called. 

can you isolate it in a simple failure case? in a simple example the callback works as I mentioned recently:
https://gist.github.com/kalikaneko/9ab5bcbb22075f9c98be


> these changes are needed for us to run the user-agent multiuser for now. Do you think these won't be needed with your latest changes? 

@bwagnerr: some changes might be needed to adapt to multiuser, I'm not sure, but review if the strategy in ab50358 would work for you (briefly: the underlying collection is what's cached, not the imap mailboxes, AND those are indexed by userid).

tuliocasagrande:
We successfully migrated from ImapAccount to Account.
You can close this PR now.

Stay awesome!

kalikaneko:
so, that change fixed the whole problem with notifications?

tuliocasagrande:
It looks like everything is OK and we have our tests green, but we'll keep an eye on that.

kalikaneko:
great to know!

PR #231, opened 2016-03-23
Bug/several mail fixes

kalikaneko:


src/leap/mail/mail.py
```diff
@@ -744,7 +746,7 @@ def __cb_signal_unread_to_ui(self, unseen):
         :param unseen: number of unseen messages.
         :type unseen: int
         """
-        emit_async(catalog.MAIL_UNREAD_MESSAGES, str(unseen))
+        emit_async(catalog.MAIL_UNREAD_MESSAGES, self.store.uuid, str(unseen))
```
meskio: That will require changes in `bitmask_client/src/leap/bitmask/gui/mail_status.py` I don't think `_mail_handle_imap_events_slot` does handle the uuid. And probably we should update the comment at `leap_pycommon/src/leap/common/events/catalog.py` describing correctly the usage of this event.

kalikaneko: mail_status was already changed

kalikaneko: common change: https://github.com/leapcode/leap_pycommon/pull/140

meskio: I don't understand how mail_status handle the uuid, but I will trust you that this is fine and go ahead with it.

PR #232, opened 2016-03-29
[feature] SMTP delivery bounces

kalikaneko:


src/leap/mail/outgoing/service.py
```diff
@@ -145,7 +149,7 @@ def send_message(self, raw, recipient):
         :return: a deferred which delivers the message when fired
         """
         d = self._maybe_encrypt_and_sign(raw, recipient)
-        d.addCallback(self._route_msg)
+        d.addCallback(self._route_msg, raw)
         d.addErrback(self.sendError)
```
meskio: Is not missing `raw` as parameter in this `addErrback`?

src/leap/mail/smtp/bounces.py
```diff
@@ -0,0 +1,78 @@
...
+I'm sorry to have to inform you that your message could not be delivered to one
+or more recipients.
+
+The reasons I got for the error are:
+
+{raw_error}
+
+--- Original message (*before* it was encrypted by bitmask) ----:
+
+{orig}"""
```
meskio: Should we mention here how to contact your own provider? We could even add the url to open a ticket in your provider https://<yourprovider>/tickets/new. Or this will make the users start spamming with tickets to often?

src/leap/mail/smtp/bounces.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def _format_msg(self, error_data, to, date, orig):
+        msg = Message()
+        msg.add_header(
+            'From', 'bitmask-bouncer@localhost (Bitmask Local Agent)')
+        msg.add_header('To', to)
+        msg.add_header('Subject', 'Undelivered Message')
+        msg.add_header('Date', date)
+        msg.set_payload(BOUNCE_TEMPLATE.format(
+            raw_error=error_data, orig=orig))
```
meskio: I'm wondering if orig should not be an attachment with the right mime type so thunderbird displays it correctly, it might be too verbose for most users to get all the headers raw there. Anyway, this can go for a second iteration if it's too much work. Maybe adding a `TODO` will suffice.

src/leap/mail/outgoing/service.py
kalikaneko: you're right, I had only added in the use below.

src/leap/mail/smtp/bounces.py
kalikaneko: I'm afraid that could be prone to overload yes. Specially since we'll be mixing together issues like network connectivity. I added a comment to the bounce message saying that user might want to contact provider if the problem persists.

src/leap/mail/smtp/bounces.py
kalikaneko: Right. There's https://tools.ietf.org/html/rfc6522 for proper structuring the delivery error reports. I added a TODO for future iterations.
FWIW, I believe gmail delivers the bounced message inline, headers and all.

PR #235, opened 2016-04-11
Remove leftover print statement

cacarrara:
The print statement only printed a number. Seeing the print you cannot
know what was printed. Seems that this line was left during a debug
process.


kalikaneko:
merged, thanks!

PR #236, opened 2016-04-14
Adds user_id into Account

cacarrara:
Previously Account used user id from the store, but this attribute is
optional and None by default. This caused the collection_mapping to be
unable to distinct between multiple users message collections.

This change adds a non optional user_id attribute to Account and use it
to index the collection_mapping.

It's related with Pixelated [bug](https://github.com/pixelated/pixelated-user-agent/issues/674) leaking email across user agent.

### DIFF
--- a/src/leap/mail/imap/account.py
+++ b/src/leap/mail/imap/account.py
@@ -60,7 +60,7 @@ class IMAPAccount(object):
 
     selected = None
 
-    def __init__(self, user_id, store, d=defer.Deferred()):
+    def __init__(self, store, user_id, d=defer.Deferred()):
         """
         Keeps track of the mailboxes and subscriptions handled by this account.
 
@@ -69,12 +69,13 @@ class IMAPAccount(object):
         You can either pass a deferred to this constructor, or use
         `callWhenReady` method.
 
+        :param store: a Soledad instance.
+        :type store: Soledad
+
         :param user_id: The identifier of the user this account belongs to
                         (user id, in the form user@provider).
         :type user_id: str
 
-        :param store: a Soledad instance.
-        :type store: Soledad
 
         :param d: a deferred that will be fired with this IMAPAccount instance
                   when the account is ready to be used.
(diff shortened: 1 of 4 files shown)


kalikaneko:
The changes look good to me.
However, I'm in favor of a strategy to avoid similar bugs in other cases, beyond this pr:

* Make user_id mandatory in soledad initialization.
* Protect the add_msg methods in mailbox.

src/leap/mail/imap/account.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, user_id, store, d=defer.Deferred()):
         # about user_id, only the client backend.
 
         self.user_id = user_id
-        self.account = Account(store, ready_cb=lambda: d.callback(self))
+        self.account = Account(store, user_id, ready_cb=lambda: d.callback(self))
```
kalikaneko: minor change: do you mind to change the order of the parameters, so that it matches the signature for IMAPAccount?

kalikaneko:
another option would be to raise an exception in the getter for userid in soledad, in case that something tries to use it without being initialized. I'm in favor of removing getter/setter boilerplate, but in this case it might be a good safety catch. what do you think?

src/leap/mail/imap/account.py
kalikaneko: nevermind, maybe it's easier to change the IMAPAccount order. I'll merge, we can discuss further changes later.

kalikaneko:
merged

cacarrara:
Hello, Kali! First of all, thanks for accepting the suggestion.

About a mandatory user id on Soledad, it's already this way today with uuid parameter. Beyond that we also found a way to improve the Soledad initialization and there is a [pull request open](https://github.com/leapcode/soledad/pull/315) for it. It would be good hear your thoughts there.

That said, maybe the exception on getter for userid on Soledad could be unnecessary, but we can continue this talk on Soledad PR.

On the second topic about user id verification on add_msg() method I confess that the logic on Account and Message Collection is a bit confused to me yet. But I was talking about it with @bwagnerr and we agreed that could be too much knowledge on MessageCollection about the store object. Maybe this check should be done and we can try to figure out a good place to it.

Still about it, there is something that is not clear to me. At first moment I agree that the user 'to' need to be the same from store. But there is any other use case where it isn't true? I don't know about aliases yet to tell if it could raise a problem.

Thanks again,

PR #238, opened 2016-04-26
Fix changelog renaming

cacarrara:
The changelog file was renamed using rst extension, but the setup.py
uses the same file without extension. So, this change revert to the
expected name format.


cacarrara:
It was fixed by this https://github.com/leapcode/leap_mail/commit/061d3f

PR #225, opened 2016-02-23
Fix verify signature on encrypted email from Apple Mail

jeffhsta:
The leap.mail has a step that decrypt an email and at the same time it should verify signature.
It works when using Thunderbird to send an email, but when using Apple Mail that step always return "could not verify signature" because of the way that Apple Mail encrypt and sign the email.

Issue: https://github.com/pixelated/pixelated-user-agent/issues/605

PR #243, opened 2016-05-25
Fix signature validation on encrypted email from Apple Mail

cacarrara:
**Disclaimer:**
As we previously talked with @kalikaneko and @jeffhsta we think could be easier to open a new pull request to fix this bug because the previous one (https://github.com/leapcode/leap_mail/pull/225) got pretty outdated. When this PR be accept the old one can be closed.

**Context:**
Leap mail has a step that uses keymanager to decrypt an email and at the same time it validates signature. It works well when the incoming email was created using Thunderbird, but when using Apple Mail that step is not enough because the way Apple Mail encrypt and sign the email. While Thunderbird does something like `gpg -se [message]` to encrypt and sign an email, Apple Mail first sign the message and only after it encrypt the previous signed message. Something like that `gpg -s [message] | gpg -e [signed_message]`. So, when keymanager decrypt a message from Apple Mail, the result is a "multipart/signed" message which still need to have its signature verified.

**Change:**
What this pull request does is add a new step after the decryption process to check if the decrypted message is a "multipart/signed", if it's true, the verify signature method is called. If it's not, the previous process continue as expected.
Besides that this pull request also adds a test to this new process flow. To do this test we added a new message file in `incoming/test` directory. This is an encrypted and signed message built by Apple Mail and we use it during the tests. Also, the way we thought to write the test was splitting the statement which adds the verified signature header to a new method, so during the test we can mock it and check it's called.

### DIFF
--- a/src/leap/mail/incoming/service.py
+++ b/src/leap/mail/incoming/service.py
@@ -458,10 +458,8 @@ class IncomingMail(Service):
                     self.LEAP_SIGNATURE_HEADER,
                     self.LEAP_SIGNATURE_INVALID)
             else:
-                decrmsg.add_header(
-                    self.LEAP_SIGNATURE_HEADER,
-                    self.LEAP_SIGNATURE_VALID,
-                    pubkey=signkey.fingerprint)
+                self._add_verified_signature_header(decrmsg,
+                                                    signkey.fingerprint)
             return decrmsg.as_string()
 
         if msg.get_content_type() == MULTIPART_ENCRYPTED:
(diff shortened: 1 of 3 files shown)


cacarrara:
Hi, @kalikaneko. We're looking forward for this PR be accepted. Let us know if you have any doubt or concerns about it.
Thanks.

src/leap/mail/incoming/service.py
```diff
@@ -458,10 +458,8 @@ def add_leap_header(ret):
                     self.LEAP_SIGNATURE_HEADER,
                     self.LEAP_SIGNATURE_INVALID)
             else:
-                decrmsg.add_header(
-                    self.LEAP_SIGNATURE_HEADER,
-                    self.LEAP_SIGNATURE_VALID,
-                    pubkey=signkey.fingerprint)
+                self._add_verified_signature_header(decrmsg,
+                                                    signkey.fingerprint)
```
cacarrara: @meskio, actually `_do_fetch` returns a list of `SoledadDocuments` which seems contains encrypted message. For sure we could check the header on a decrypted message, but for now we cannot see how to get these decrypted messages after the `_do_fetch` call. Do you have any suggestion?

Here an example of `_do_fetch()` return: http://paste.debian.net/726745/

meskio: Ok, then I'll merge it as it is, I don't have good sugestions here.

PR #245, opened 2016-06-17
[bug] initialize OpenSSL context just once in leap.mail

mnandri:
Do not initialize the openssl context on each call to get mail payload phash.

The openSSL backend should only be initialized once because it is activating the os random engine
which in turn unregister and free the current engine first. This is very tricky when operations are running in threads as it  momentarily unregister the openssl crypto callbacks that makes openssl thread safe.

- Resolves: #8180 with the soledad PR #324


kalikaneko:
rebased and merged

PR #247, opened 2016-07-24
[refactor] deprecating u1db, using l2db instead.

mnandri:
To keep compatibility with soledad upgrades. It will namely cause revision conflicts
to not be properly handled otherwise.


kalikaneko:
merged

mnandri:
¡muchas gracias!

